PR #7, opened 2022-12-24
Add vex command

luhring:
- Adds new `wolfictl vex` command to generate a VEX document for a Wolfi package
- Adds a new library function `vex.FromPackageConfiguration` that takes a parsed Melange configuration and generates the VEX data for that package (this is used to drive the `wolfictl vex` command)


pkg/vex/vex.go
```diff
@@ -0,0 +1,113 @@
...
+		return vex.StatusNotAffected
+	}
+
+	return vex.StatusFixed
+}
+
+func generateDocumentID(packageName string) string {
+	uuid := uuid.New()
+	return fmt.Sprintf("vex-%s-%s", packageName, uuid)
+}
```
puerco, line 113: We need to make this ID deterministic. But let's merge it now as it is and we can improve it in a follow up

luhring: Good call

PR #8, opened 2022-12-27
Add golangci-lint job, fix errors and bugs

puerco:
This PR adds golangci-lint into the repository and fixes all errors in the code. It fixes some minor bugs and starts to remove deprecated code from some modules. It also adds a new job to run golangci-lint in a presubmit.

Signed-off-by: Adolfo García Veytia (Puerco) <puerco@chainguard.dev>

### DIFF
--- a/pkg/lint/types.go
+++ b/pkg/lint/types.go
@@ -31,7 +31,7 @@ type Rule struct {
 	// Severity is the severity of the rule.
 	Severity Severity
 
-	// Function is the function that lints a single configuration.
+	// LintFunc is the function that lints a single configuration.
 	LintFunc Function
 
 	// ConditionFuncs is a list of and-conditioned functions that check if the rule should be executed.
--- a/pkg/lint/types.go
+++ b/pkg/lint/types.go
@@ -5,8 +5,8 @@ import (
 	"github.com/hashicorp/go-multierror"
 )
 
-// LintFunc is a function that lints a single configuration.
-type LintFunc func(build.Configuration) error
+// Function is a function that lints a single configuration.
+type Function func(build.Configuration) error
 
 // ConditionFunc is a function that checks if a rule should be executed.
 type ConditionFunc func() bool
@@ -31,8 +31,8 @@ type Rule struct {
 	// Severity is the severity of the rule.
 	Severity Severity
 
-	// LintFunc is the function that lints a single configuration.
-	LintFunc LintFunc
+	// Function is the function that lints a single configuration.
+	LintFunc Function
 
 	// ConditionFuncs is a list of and-conditioned functions that check if the rule should be executed.
 	ConditionFuncs []ConditionFunc
--- a/pkg/update/releaseMonitor.go
+++ b/pkg/update/releaseMonitor.go
@@ -32,13 +32,14 @@ const (
 	releaseMonitor    = "RELEASE_MONITOR"
 )
 
-func (m MonitorService) getLatestReleaseMonitorVersions(mapperData map[string]Row, melangePackages map[string]build.Configuration) (map[string]string, []string, error) {
-	var errorMessages []string
-	packagesToUpdate := make(map[string]string)
-	// iterate packages from the target git repo and check if a new version is available
-	for _, config := range melangePackages {
+func (m MonitorService) getLatestReleaseMonitorVersions(
+	mapperData map[string]Row, melangePackages map[string]build.Configuration,
+) (packagesToUpdate map[string]string, errorMessages []string, err error) {
+	packagesToUpdate = make(map[string]string)
 
-		item := mapperData[config.Package.Name]
+	// iterate packages from the target git repo and check if a new version is available
+	for i := range melangePackages {
+		item := mapperData[melangePackages[i].Package.Name]
 		if item.Identifier == "" {
 			continue
 		}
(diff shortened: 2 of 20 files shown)


pkg/lint/types.go
```diff
@@ -31,8 +31,8 @@ type Rule struct {
 	// Severity is the severity of the rule.
 	Severity Severity
 
-	// LintFunc is the function that lints a single configuration.
-	LintFunc LintFunc
+	// Function is the function that lints a single configuration.
+	LintFunc Function
```
luhring, line 35: Was the doc comment change intentional?

pkg/update/releaseMonitor.go
```diff
@@ -32,13 +32,14 @@ const (
...
-	for _, config := range melangePackages {
+func (m MonitorService) getLatestReleaseMonitorVersions(
+	mapperData map[string]Row, melangePackages map[string]build.Configuration,
+) (packagesToUpdate map[string]string, errorMessages []string, err error) {
+	packagesToUpdate = make(map[string]string)
 
-		item := mapperData[config.Package.Name]
+	// iterate packages from the target git repo and check if a new version is available
+	for i := range melangePackages {
+		item := mapperData[melangePackages[i].Package.Name]
```
luhring, line 42: Why change to using index integers instead of using the slice elements directly?

pkg/vex/vex.go
```diff
@@ -34,29 +34,28 @@ func FromPackageConfiguration(buildCfg build.Configuration, vexCfg Config) (vex.
...
 	secfixesStatements := statementsFromSecfixes(cfg.Secfixes, purls)
 	advisoriesStatements := statementsFromAdvisories(cfg.Advisories, purls)
 
 	// don't include "not_affected" statements from secfixes that are obviated by statements from advisories
 	notAffectedVulns := make(map[string]struct{})
-	for _, stmt := range advisoriesStatements {
-		if stmt.Status == vex.StatusNotAffected {
-			notAffectedVulns[stmt.Vulnerability] = struct{}{}
+	for i := range advisoriesStatements {
+		if advisoriesStatements[i].Status == vex.StatusNotAffected {
```
luhring, line 45: Why change to using index integers instead of using the slice elements directly?

pkg/vex/vex.go
```diff
@@ -34,29 +34,28 @@ func FromPackageConfiguration(buildCfg build.Configuration, vexCfg Config) (vex.
...
+			notAffectedVulns[advisoriesStatements[i].Vulnerability] = struct{}{}
 		}
 	}
 	var statements []vex.Statement
-	for _, sfStmt := range secfixesStatements {
-		if _, seen := notAffectedVulns[sfStmt.Vulnerability]; !seen {
-			statements = append(statements, sfStmt)
+	for i := range secfixesStatements {
+		if _, seen := notAffectedVulns[secfixesStatements[i].Vulnerability]; !seen {
```
luhring, line 51: Why change to using index integers instead of using the slice elements directly?

pkg/vex/vex.go
```diff
@@ -65,15 +64,17 @@ func statementsFromAdvisories(advisories build.Advisories, purls []string) []vex
 	var stmts []vex.Statement
 
 	for v, entries := range advisories {
-		for _, entry := range entries {
-			stmts = append(stmts, statementFromAdvisoryContent(entry, v, purls))
+		for i := range entries {
+			stmts = append(stmts, statementFromAdvisoryContent(&entries[i], v, purls))
```
luhring, line 68: Why change to using index integers instead of using the slice elements directly?

pkg/vex/vex.go
```diff
@@ -15,7 +15,7 @@ type Config struct {
 }
 
 // FromPackageConfiguration generates a new VEX document for the Wolfi package described by the build.Configuration.
-func FromPackageConfiguration(buildCfg build.Configuration, vexCfg Config) (vex.VEX, error) {
+func FromPackageConfiguration(buildCfg *build.Configuration, vexCfg Config) (vex.VEX, error) {
```
luhring, line 18: Why introduce a pointer here?

pkg/update/releaseMonitor.go
puerco: This is mainly to avoid large allocations caused by copying the structs on each iteration.

pkg/vex/vex.go
puerco: Same, moving copies of these structs around will get flagged by the linter.

pkg/lint/types.go
puerco: No, good catch :)

pkg/update/releaseMonitor.go
luhring: Okay, hmm. We can move forward for now but let's think about when following this kind of lint rule has more benefit vs. cost to the code base, particularly in terms of readability and pointer complexity

pkg/vex/vex.go
luhring: Okay, ditto [this comment](https://github.com/wolfi-dev/wolfictl/pull/8#discussion_r1058645226) — e.g. now this function can panic during a nil pointer dereference

luhring (approved):
Approving — but let's just keep an eye on the effects of applying some of these lint rules 😃

PR #9, opened 2022-12-28
Make VEX doc generation deterministic

luhring:
This adjusts `wolfictl vex ...` to produce VEX documents in a **deterministic** way.

Document IDs are now produced as a function of the package name and the last git commit that affect the package configuration file.

Document timestamps are now set to the author timestamp for the last git commit that affect the package configuration file.


luhring:
Talked w/ @puerco, and he'll incorporate this into #11 .

PR #10, opened 2022-12-28
Lint cleanup

luhring:
Signed-off-by: Dan Luhring <dluhring@chainguard.dev>

### DIFF
--- a/pkg/lint/rules.go
+++ b/pkg/lint/rules.go
@@ -10,19 +10,6 @@ import (
 	"chainguard.dev/melange/pkg/build"
 )
 
-var (
-	reValidSHA256 = regexp.MustCompile(`^[a-fA-F0-9]{64}$`)
-	reValidSHA512 = regexp.MustCompile(`^[a-fA-F0-9]{128}$`)
-
-	forbiddenRepositories = []string{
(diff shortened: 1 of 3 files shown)


pkg/lint/rules.go
```diff
@@ -10,19 +10,6 @@ import (
...
-
-	forbiddenRepositories = []string{
-		"https://packages.wolfi.dev/os",
-	}
-
-	forbiddenKeyrings = []string{
-		"https://packages.wolfi.dev/os/wolfi-signing.rsa.pub",
-	}
-)
-
```
rawlingsj, line 25: I don't have an opinion on moving these from the top of the file to the individual functions so happy to merge but did wonder if there was a specific reason to change from the original style?

luhring: We can leave them if that ends up seeming better! The specific reason is just to minimize the scope for each of these things. These variables each seemed to go with individual rules, so keeping them logically close to the rules and not exposing them beyond the rule definition made sense in my head. E.g. I couldn't think of any other rules where we'd use `forbiddenRepositories` other than the existing rule that prevents the use of forbidden repositories. But happy to revert if you'd like.

luhring:
I'll add these changes to a future PR

PR #18, opened 2023-01-03
Add advisory commands

luhring:
This PR adds the new `advisory ...` commands to `wolfictl`, which are used to interact with **advisory data** (a new concept that will eventually replace _"secfixes"_) in Wolfi.

- `wolfictl advisory create ...`
- `wolfictl advisory update ...`
- `wolfictl advisory list ...`
- `wolfictl advisory sync-secfixes ...`

This PR also adds a new `configs` package for operating on sets of package config files at a time, especially useful for distro-wide logic. The idea here is that you start by creating a new `configs.Index`, and then you can find and mutate configs in the index in a way that's type-safe and scalable.


jdolitsky (approved):
do it!

PR #21, opened 2023-01-05
Fix doc comment glitch from automatic lint fixes

luhring:
Signed-off-by: Dan Luhring <dluhring@chainguard.dev>


luhring:
Overriding since it's a trivial change

PR #23, opened 2023-01-07
Add wolfictl bump subcommand

puerco:
From the help text:
```
wolfictl bump increases the epoch field in melange configuration files

The bump subcommand allows increments version numbers in packages. For now it will only bump epoch numbers but a future version will allow users to control versions expressed in semver.

wolfictl bump can take a filename, a package or a file glob, increasing the version in each matching configuration file
```

This works by doing a string replacement on the epoch field in the yaml configs. It does no re-marshaling at all as it was causing a huge diff.

Signed-off-by: Adolfo García Veytia (Puerco) <puerco@chainguard.dev>

### DIFF
--- a/pkg/cli/bump.go
+++ b/pkg/cli/bump.go
@@ -1,0 +1,141 @@
+package cli
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"chainguard.dev/melange/pkg/build"
+	"github.com/spf13/cobra"
+)
+
+const epochPattern = `epoch: %d`
+
+type bumpOptions struct {
+	repoDir string
+	epoch   bool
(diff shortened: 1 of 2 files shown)


rawlingsj (approved):
It might be worth us following up with a test so folks can extend this.  Another similar usecase  is to request downstream packages have their epoch bumped after a new version update.  I.e. when `go` is updated we need to bump other packages so they are rebuilt, here's a manually created example of that https://github.com/wolfi-dev/os/pull/176/files

All stuff we can add to this though.



There's been a few issues lately automating the update of melange configs

rawlingsj:
FWIW related to the above comment, we should be able to use [dag](https://github.com/wolfi-dev/dag) to figure out the downstream packages that need their epoch bumped.

PR #27, opened 2023-01-17
Add command to discover potential vulnerabilities for packages

luhring:
This adds a new command: `wolfictl advisory discover`.

The `discover` command searches the [Wolfi secfixes-tracker](https://secfixes-tracker-nd2dq3gc7a-uk.a.run.app/) to find potential matches of Wolfi packages to CVEs. The command automatically adds new advisories to any package config file where the CVE didn't already have an advisory, and it initializes the new advisory with the status `under_investigation`.

Like the other `advisory` commands, `discover` expects to be run in the root of the `wolfi-dev/os` repo. It can also take one or more paths to Wolfi package configs.

This PR also replaces the `yaml.Encoder` with the new `formatted.Encoder` from [yam](https://github.com/chainguard-dev/yam). So you'll start to see gap lines added back into our YAML files.

---

### Updates to the `configs` API

To make it easier to operate on Wolfi YAML configs, this PR also makes progress on refining the API in the `configs` Go package.

Given an `index` of configs, you can now operate on any section of the config file. And you can do this for all configs or only a subset of configs.

For example, to update the advisories section of the "curl" package's config:

```go
_ = index.Select().WherePackageName("curl").UpdateAdvisories(func(cfg build.Configuration) (build.Advisories, error) {
    advisories := cfg.Advisories

    // do stuff to this `advisories` struct

    return advisories, nil
})
```

You can also map packages to any data shape needed. For example, to map all configs to a string that reports the package's name and version:

```go
items, _ := configs.Map(index.Select(), func(e configs.Entry) (string, error) {
    pkg := e.Configuration().Package
    return fmt.Sprintf("%s %s-r%s", pkg.Name, pkg.Version, pkg.Epoch), nil
})
```


pkg/advisory/advisory.go
```diff
@@ -140,89 +128,62 @@ func Latest(entries []build.AdvisoryContent) *build.AdvisoryContent {
...
+			if !m.CPE.VersionRange.Includes(cfg.Package.Version) {
+				continue
+			}
+
+			vulnID := m.Vulnerability.ID
+			_, exists := cfg.Advisories[vulnID]
+
+			if exists {
+				// TODO: Should we allow for updating existing advisories if previously read vuln
+				// data has been updated?
```
jdolitsky, line 167: cant help but think a log line here will help later, but maybe its too verbose

pkg/cli/advisory_discover.go
```diff
@@ -0,0 +1,40 @@
...
+		Short:         "search for new potential vulnerabilities and create advisories for them",
+		SilenceErrors: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			index, err := newConfigIndexFromArgs(args...)
+			if err != nil {
+				return err
+			}
+
+			secfixesTracker := sftracker.New(
+				"secfixes-tracker-nd2dq3gc7a-uk.a.run.app",
```
jdolitsky, line 23: constant/option plz

jdolitsky (approved):
damn

pkg/configs/updaters.go
```diff
@@ -0,0 +1,234 @@
+package configs
```
rawlingsj, line 1: Nice!  I wonder at some point it might be nice to have a test or two as we'll probably want to reuse this for future automation, so a test would make it easier to extend?  We've got some unit tests in this repo now so maybe we could reuse the approach?

pkg/advisory/advisory.go
luhring: Nice — added in becec34242f0616adea99e55ab1cac0b060225c4!

pkg/cli/advisory_discover.go
luhring: Great thinking, added in 899f590ad2386bdb0c8f43d2cb9bbd059842106d

jdolitsky (approved):

pkg/configs/updaters.go
rawlingsj: awesome sounds good to me!

PR #31, opened 2023-01-24
Support updating gitmodules + handle melange config files that live in subfolders, not just the root directory

rawlingsj:


rawlingsj:
This also improved the github graphql tests as it was a little hard to extend

imjasonh (approved):

PR #41, opened 2023-02-08
Add generate-index which generates an index from apks in a bucket

imjasonh:
This uses @jdolitsky's fast magic so that the index is generated in <5 minutes:

```
wolfictl generate-index
```

This runs against the prod wolfi bucket by default, but also supports `--bucket=stage3` for example, or just `--bucket=gs://blah/blah`.

If `--signing-key` is passed, the APKINDEX is signed with that key. If `--publish` is passed, the APKINDEX is written back to the bucket. To publish, you must also sign.

This also adds support for `wolfictl index --repo=APKINDEX.tar.gz` (an unfortunate misnomer) for local testing.

Progress toward https://github.com/wolfi-dev/os/issues/551


pkg/cli/apk.go
```diff
@@ -100,3 +120,174 @@ func Index() *cobra.Command {
...
+				k, err := os.Open(signingKey)
+				if err != nil {
+					return err
+				}
+				defer k.Close()
+			}
+
+			idx := &repository.ApkIndex{
+				// TODO: sign the index.
+				// https://github.com/chainguard-dev/melange/blob/main/internal/sign/sign.go#L34
```
jdolitsky, line 168: is it not done below?

jdolitsky (dismissed):

pkg/cli/apk.go
imjasonh: It's true! Removed the TODO.

I also removed the description here, since it wasn't adding much. Our current index is generated without a description.

kaniini (dismissed):

pkg/cli/apk.go
```diff
@@ -4,15 +4,28 @@ import (
...
+	"log"
 	"net/http"
 	"os"
+	"path"
+	"runtime"
+	"sort"
 	"strings"
+	"sync"
 
+<<<<<<< HEAD
```
kaniini, line 16: unfixed merge conflict?

imjasonh: Mortifying. Fixed!

jdolitsky (approved):

kaniini (approved):

kaniini (approved):

jdolitsky (approved):

PR #44, opened 2023-02-09
move dag commands into wolfictl

imjasonh:
This adds `wolfictl svg`, `wolfictl text` and `wolfictl pod` copied wholesale from https://github.com/wolfi-dev/dag.

Linting is more strict here than in that repo, so there's a [second commit](https://github.com/wolfi-dev/wolfictl/pull/44/commits/85f9490cad23358817140827393da738b6541c08) that fixes up all the lint findings, and ignores/disables some.

After this we should update workflows that install and run `dag` to install and run `wolfictl` instead.

@rawlingsj if you think there are more `dag` code changes likely in the near term, we can also just put this off for a while.

Signed-off-by: Jason Hall <jason@chainguard.dev>


imjasonh:
Dag code depends on graphviz, which requires cgo. I've removed `CGO_ENABLED=0` from the Makefile, and all the unused/broken ko and goreleaser stuff from there as well. We don't plan to produce an image for wolfictl (AFAIK), or do traditional releases of the tool.

rawlingsj:
Not sure if this matters but I have been using an image of wolfictl from here https://github.com/wolfi-dev/tools.  Again it may not be relevent but wolfi has graphviz? https://github.com/wolfi-dev/os/blob/main/graphviz.yaml

imjasonh:
> Not sure if this matters but I have been using an image of wolfictl from here https://github.com/wolfi-dev/tools. Again it may not be relevent but wolfi has graphviz? https://github.com/wolfi-dev/os/blob/main/graphviz.yaml

Hmm interesting. The library we use explicitly doesn't require graphviz to be installed (https://github.com/goccy/go-graphviz#features) but it seems to achieve that by effectively building graphviz into the binary using cgo. We actually use Scott's fork which drops everything but svg support: https://github.com/goccy/go-graphviz/commit/fd4faf331d281482d67119caa4192b85c9a663dc

So it sounds like we'll want the wolfictl image to include glibc and depend on gcc at build-time to be able to build wolfictl. I can try that out. _Or_, I can make this code not use graphviz-the-library at all, and instead just generate .dot and pipe it to `graphviz`-the-tool, and make the wolfictl image install graphviz.

rawlingsj:
Ah cool, sounds interesting!  No idea which one's going to be easiest though :)

luhring (dismissed):
Love this direction! 🚀

jdolitsky (approved):

PR #56, opened 2023-02-13
wolfictl update, fix regex to handle CVE names with hyphen separators

rawlingsj:
Signed-off-by: James Rawlings <jrawlings@chainguard.dev>


kaniini (approved):

luhring:
I would even suggest that CVE identifiers **must** be in the normal/correct form for clarity — e.g. `CVE-2000-1234`

PR #58, opened 2023-02-14
wolfictl lint: allow users to skip rules via a CLI flag

rawlingsj:
This also contains a fix to not be so strict when checking if a yaml file is indeed a melange config.  We are seeing linting being applied to files that are not melange configs, this assumes a bare minimum that package name and package version are set, to indicate a yaml file should be linted.


cpanato (approved):
lgtm

PR #67, opened 2023-02-17
wolfictl pod: add --public-key-bucket to optionally override the buck…

rawlingsj:
…et location of public key

This is useful if we want to use sub folders in the source bucket but they all use the same signing pub and private key

### DIFF
--- a/pkg/cli/pod.go
+++ b/pkg/cli/pod.go
@@ -45,9 +45,10 @@ func gcloudProjectID(ctx context.Context) (string, error) {
 }
 
 func cmdPod() *cobra.Command {
-	var dir, arch, project, bundleRepo, ns, cpu, ram, sa, sdkimg, cachedig, bucket, srcBucket, signingKeyName string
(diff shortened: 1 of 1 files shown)


rawlingsj:
@imjasonh could you take a look at this when you get some time pls and see what you think

pkg/cli/pod.go
```diff
@@ -145,11 +150,12 @@ set -euo pipefail
 # Download all packages so we can avoid rebuilding them.
 mkdir -p ./packages/{{.arch}}
 gsutil -m rsync -r {{.bucket}}{{.arch}} ./packages/{{.arch}} || true
-gsutil cp {{.bucket}}{{.signingKeyName}}.rsa.pub .
+gsutil cp {{.publicKeyBucket}}{{.signingKeyName}}.rsa.pub .
```
imjasonh, line 153: If you do --signing-key=foo/bar/key.rsa.pub does that work?

rawlingsj: Unfortunately not because the public key is in the folder above

imjasonh (approved):

cpanato (approved):

PR #81, opened 2023-02-27
lint: require tag and expected-commit for git-checkout

imjasonh:
All packages in wolfi-dev/os should have these fields, except ruby yajl, which is accounted for in https://github.com/wolfi-dev/os/pull/714

Cloning a branch or tag without checking the expected-commit leaves us open to moving branches/tags resulting in unexpected changes being built and released.


imjasonh:
This is now rebased on https://github.com/wolfi-dev/wolfictl/pull/88

luhring (approved):

PR #83, opened 2023-02-28
add `wolfictl make` to execute `make` for us

imjasonh:
Request for feedback, with some remaining TODOs to fix graph generation to handle `provides:`es.

Anyway, the idea is that `wolfictl make` will generate a DAG of targets and build them in order, invoking `make packages/aarch64/helm-3.11.1-r1.apk`, etc.

This wouldn't let us get rid of the Makefile completely, but it _would_ let us drop the hand-curated list of targets which must be kept in order.

### DIFF
--- a/pkg/cli/commands.go
+++ b/pkg/cli/commands.go
@@ -25,6 +25,7 @@ func New() *cobra.Command {
 		cmdPod(),
 		cmdSVG(),
 		cmdText(),
+		cmdMake(),
 	)
 
 	return cmd
--- a/pkg/cli/make.go
+++ b/pkg/cli/make.go
@@ -1,0 +1,56 @@
+package cli
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+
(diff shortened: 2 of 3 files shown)


imjasonh:
The wrinkle we hit here is that some packages have `provides: ['go=1.19.999']` or `provides: ['go=1.20.999']`, and things have a dependency on `'go~=1.19'`. The current dumb version of this code just chops off the second part in both cases, and something that depends on `go~=1.19` will randomly choose a single one of either `go-1.19` or `go-1.20`. This is _probably_ fine for now, since both of those have the same dependencies.

A better smarter version of this would be to better match the version.

The result at least for now is that there's a topo-sorted set of `make` invocations that _should_ be functionally equivalent to the hand-curated list in `Makefile` and `make all`.

To roll this out we'd switch `wolfictl pod` to (_somehow_) get its own `wolfictl` binary into the container, so it can run `wolfictl make` instead of `make all` like it does today.

luhring (approved):
LGTM! one nit

PR #93, opened 2023-03-06
add a new checks subcommands and so name check for wolfi CI

rawlingsj:


rawlingsj:
@kaniini you'll probably want to take a look at this one

pkg/checks/so_name.go
```diff
@@ -0,0 +1,279 @@
...
+	// Writer the body to file
+	_, err = io.Copy(out, resp.Body)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+//nolint:gocritic // hugeParam for entry
+func (o SoNameOptions) getSonameFiles(dir string) ([]string, error) {
```
kaniini, line 235: Why not use a pointer to avoid the hugeParam issue?

kaniini:
Consider the following scenario:

Old package:

libfoo.so.1 is symlinked to libfoo.so.1.2.3, `DT_SONAME` is `libfoo.so.1`.

New package:

libfoo.so.1 is symlinked to libfoo.so.1.2.4, `DT_SONAME` remains `libfoo.so.1`.

Will this code be correct?  I don't see any tests covering this particular scenario.  It may be helpful to use `debug/elf` here to pull the `DT_SONAME`, like Melange does.

pkg/checks/so_name.go
rawlingsj: only becuase by IDE complained with a warning when making it a pointer
```
Struct SoNameOptions has methods on both value and pointer receivers. Such usage is not recommended by the Go Documentation.
```
I've not looked closely at why, so probably my code.  Will check in the morning.

rawlingsj:
> Consider the following scenario:

Ok thanks, yeah that's a good one, will add a test to cover that.

rawlingsj:
@kaniini I've pushed a commit related to `DT_SONAME` which I found in [melange](https://github.com/chainguard-dev/melange/blob/f0e9e0df807d9dcc83b55dd09f4294502a5fdceb/pkg/build/package.go#L451).  Though I have to admit I'm not 100% sure I know what I'm doing for this one :)

AFAICT ABI breaking changes only apply to major version increments of the soname file.  Granted there's probably more reliable sources of info but was basing this from https://en.wikipedia.org/wiki/Soname.

So I've added a few more tests since the original commit yesterday.  To make sure we only fail on major version increments AND to include symlinks.  But I may have got the latter wrong.

kaniini:
This appears to look okay, but it needs the merge conflict sorted.

rawlingsj:
awesome, thanks for that - I was holding off rebasing in case I had to do it a few times, lemme push now

luhring (approved):

PR #135, opened 2023-04-03
wolfi update: get the git commit associated with a tag when updating …

rawlingsj:
…new versions so we can set expected-commit value in a melange config


rawlingsj:
depends on https://github.com/chainguard-dev/melange/pull/367

PR #139, opened 2023-04-03
lint calls make list and checks package names

deitch:
With wolfi-dev/os now changed to use packages in a separate file, rather than in the Makefile, `wolfictl lint` reads the wrong thing and fails.

This PR updates it to call `make list`, using that as an API to report all known packages, and finds them that way.


imjasonh (dismissed):

rawlingsj (approved):

deitch:
Thanks @rawlingsj

PR #140, opened 2023-04-04
wolfictl check update: verify update config actually does fetch or gi…

rawlingsj:
…t-checkout sources

this will help verify update config before creating a PR or in CI before merging


luhring (approved):
Nice!

PR #150, opened 2023-04-09
wolfi update: fix duplicate PRs being created, need to handle github …

rawlingsj:
…pagination when listing issues and pull requests

this also moves some calls to github so we cut down down on the number of requests made.


pkg/gh/github.go
```diff
@@ -86,6 +63,41 @@ func (o GitOptions) handleRateLimit(action func() (*github.Response, error)) err
 	return nil
 }
 
+func (o GitOptions) handleRateLimitList(action func(opt *github.ListOptions) (*github.Response, error)) error {
+	opt := &github.ListOptions{
+		PerPage: 30,
+	}
+
+	for {
+		resp, err := action(opt)
```
Dentrax, line 72: `err` check

pkg/gh/issues.go
```diff
@@ -13,17 +13,35 @@ type Issues struct {
...
 
+func (o GitOptions) ListIssues(ctx context.Context, owner, repo, state string) ([]*github.Issue, error) {
+	var openIssues []*github.Issue
+
+	err := o.handleRateLimitList(func(opt *github.ListOptions) (*github.Response, error) {
+		ilo := github.IssueListByRepoOptions{
+			State:       state,
+			ListOptions: *opt,
+		}
+		issues, resp, err := o.GithubClient.Issues.ListByRepo(ctx, owner, repo, &ilo)
```
Dentrax, line 27: You silently ignored the `err` here. [ListByRepo](https://github.com/google/go-github/blob/521fd0d1f06050e4fe370eb5eaa02a10d9e50baa/github/issues.go#L222-L244) returns `nil` on any error case. But I think this works since current code base already using in this way.

pkg/gh/github.go
```diff
@@ -86,6 +63,41 @@ func (o GitOptions) handleRateLimit(action func() (*github.Response, error)) err
...
+		if resp.StatusCode == http.StatusUnauthorized {
+			return errors.Wrap(err, "failed to auth with GitHub, does your personal access token have the repo scope? https://github.com/settings/tokens/new?scopes=repo")
+		}
+
+		if err != nil {
+			if githubErr := github.CheckResponse(resp.Response); githubErr != nil {
+				isRateLimited, delay := o.checkRateLimiting(githubErr)
+				if isRateLimited {
+					o.Logger.Printf("retrying again later with %v second delay due to secondary rate limiting.", delay.Seconds())
+					time.Sleep(delay)
```
Dentrax, line 83: Did you mean `Ratelimiter.Wait()` here?

pkg/gh/github.go
```diff
@@ -86,6 +63,41 @@ func (o GitOptions) handleRateLimit(action func() (*github.Response, error)) err
...
+
+	for {
+		resp, err := action(opt)
+
+		if resp.StatusCode == http.StatusUnauthorized {
+			return errors.Wrap(err, "failed to auth with GitHub, does your personal access token have the repo scope? https://github.com/settings/tokens/new?scopes=repo")
+		}
+
+		if err != nil {
+			if githubErr := github.CheckResponse(resp.Response); githubErr != nil {
```
Dentrax, line 79: Maybe worth to create a new separate func for this one?

pkg/gh/github.go
rawlingsj: The err check is being done after the unauthorised status check.

pkg/gh/github.go
rawlingsj: I've been using `time.Sleep(delay)` and has been working fine, happy to keep it TBH

pkg/gh/issues.go
rawlingsj: the err is returned on line 29 unless I miss understand the comment here?

pkg/gh/github.go
rawlingsj: thanks for the feedback - updated

pkg/gh/github.go
rawlingsj: I've reworked the error handling to be clearer so think this is better now

pkg/gh/issues.go
rawlingsj: going to close this but happy to followup if there's something missed.

amouat (approved):

PR #151, opened 2023-04-11
restructure DAG graph creation and fix some errors

deitch:
This is a first attempt to update the DAG, so it can be used reliably to generate the list of packages to build.

The eventual goals are:

1. eliminate the manually-maintained list of packages in wolf-dev/os
2. use this DAG to build all of the packages in wolf-dev/os in order, automatically, without any caches and without any dependency failures

We will use the second step above to generate a regular CI test that builds them all from HEAD at least once a day

## What this PR does

* fixes a bug wherein the key to a provides actually is the filename, rather than the name of the provides (oops)
* extracts a number of repetitive sections into functions
* removes some methods that were on `Graph` and are about printing text, and moves them to a more appropriate place
* removes some repetitiveness from some of the cli output methods (text, make, etc.)
* handles (some of) multiple versions, whereas previously we would either lose information or have conflicts
* separates parsing of the yamls in the directory and building configuration structs out of them, as well as reading all of their subpackages and provides, from constructing the actual graph
* simplifies much of the building of subgraphs (from leaves and from roots) based on the separation of the previous step

The separation into `Packages` and `Graph` is particularly important. It makes it much easier to reason about what it is doing. It makes it possible to work with them separately, pulling from the appropriate location.

It also makes it clearer where we have redundancy. At a later stage, we may be able to remove some of this entirely, replacing it with other parts of this package or some of the features in apko.

## What is left to do?

A number of things

* add `Packages` and `Graph` unit tests - there were none before on `Graph`, but that doesn't mean we shouldn't add some. I don't think it should hold us up, but we should at least start.
* handle versions in the graph - the graph right now is entirely based on package/subpkg/provide names. Period. That ignores the fact that there can be and are versions. `Packages` handles it, `Graph` does not. This also should not hold us up, as we do not handle versions now, and this does not lead to duplication. But we should get it right a bit later.
* versioned requirements - we don't have many of these, just one for now. `exim.yaml` depends upon `openssl-dev>3`, which is treated as a straight-out package name. The logic for this already exists in apko, but is mostly not exported. We should leverage it. Again, this fails now, so this is not a regression. We should not hold up because of it.

I will try to get the above done, but if people feel this is ready, let's not hold it up. If this is a real improvement, let's accept it and iterate.


luhring (changes requested):
I'll review more very soon! But want to make sure we talk through getting tests in place before merging this.

tuananh:
This is nice. It feels weird that we have DAG but still maintaining an ordered list in wolfi-dev/os :)

deitch:
This is strange. It is failing in setting up the lint action itself?

pkg/dag/graph.go
```diff
@@ -29,170 +24,45 @@ func newGraph() graph.Graph[string, string] {
...
-	g := newGraph()
-
-	var packages []string
-	configs := make(map[string]build.Configuration)
-
-	err := fs.WalkDir(dirFS, ".", func(path string, d fs.DirEntry, err error) error {
-		if err != nil {
-			return err
-		}
+func NewGraph(pkgs *Packages) (*Graph, error) {
```
luhring, line 27: Let's update the doc comment for this function to keep it accurate

pkg/dag/packages.go
```diff
@@ -0,0 +1,222 @@
...
+	// sort for deterministic output
+	sort.Strings(allPackages)
+	return allPackages
+}
+
+// Sub returns a new Graph whose members are the named packages or provides that are listed.
+// If a listed element is a provides, automatically includes the package that provides it.
+// If a listed element is a subpackage, automatically includes the package that contains it.
+// If a listed element does not exist, returns an error.
+func (p Packages) Sub(names ...string) (*Packages, error) {
```
luhring, line 192: I like this function. I stumbled over the name for a second, debating if this referred to subpackages or subgraphs. (My guess is the latter?) Could we update the name for clarity?

pkg/dag/packages.go
```diff
@@ -0,0 +1,222 @@
...
+	if c, ok := p.configs[name]; ok {
+		list := make([]*Configuration, 0, len(c))
+		list = append(list, c...)
+		return list
+	}
+
+	return nil
+}
+
+// PkgInfo returns the buildp.Package struct
```
luhring, line 162: nit: sp
```suggestion
// PkgInfo returns the build.Package struct
```

pkg/dag/packages.go
```diff
@@ -0,0 +1,222 @@
...
+)
+
+// Configuration represents a configuration along with the file that sourced it
+type Configuration struct {
+	*build.Configuration
+	Path string
+}
+
+type Packages struct {
+	configs  map[string][]*Configuration
```
luhring, line 21: Question: What's the motivation for having a slice of pointers to `Configuration` instead of just `Configuration` itself? What would it mean for a slice element to be `nil`?

pkg/dag/packages.go
```diff
@@ -0,0 +1,222 @@
...
+	c := config[0]
+	if pkgName != c.Package.Name {
+		return nil, nil
+	}
+	return &c.Package, nil
+}
+
+// Packages returns a slice of the names of all packages, sorted alphabetically.
+func (p Packages) Packages() []string {
+	allPackages := make([]string, 0)
```
luhring, line 178: nit: We can set the capacity to `len(p.packages)` here if the goal is to avoid allocations

pkg/dag/packages.go
```diff
@@ -0,0 +1,222 @@
...
+	"chainguard.dev/melange/pkg/build"
+)
+
+// Configuration represents a configuration along with the file that sourced it
+type Configuration struct {
+	*build.Configuration
+	Path string
+}
+
+type Packages struct {
```
luhring, line 20: This seems like an important concept in the overhauled design, so it should have a good doc comment.

luhring (changes requested):
Overall nice work! I'm hopeful about what this will let us do in the near future. 🚀 

Re: testing, now's a good time to add some tests, particularly for `Graph` and `Packages`. It will help clarify what the expected behavior of these types is. And it will help us prove to ourselves that the bugs we're saying are fixed truly are.

pkg/dag/graph.go
deitch: Updated

pkg/dag/packages.go
deitch: Done

pkg/dag/packages.go
deitch: Added

luhring: Awesome comment!

pkg/dag/packages.go
luhring: I'm not quite following yet, maybe it was a poorly phrased question on my part. Why `map[string][]*Configuration` instead of `map[string][]Configuration`?

pkg/dag/packages.go
luhring: Okay got it. The docs say `Sub` returns a new `Graph` — but it looks like it's returning a new `*Packages`. And it looks like this populates the `Packages` object with more than just "subpackages". Maybe `SubSelection` or something would be better here?

deitch: Updated. I really missed that one. Thanks for that @luhring

deitch:
Added tests for the `Packages`, moving on to `Graph`.

deitch:
@luhring I saw you added some basic smoke test for `NewGraph()`. I disabled it temporarily in this PR, as it conflicts with some of my more contrived tests. But I will reenable it before we merge this in.

kaniini (approved):

deitch:
Please don't merge yet. Logic is not done.

luhring:
> Please don't merge yet. Logic is not done.

Shall we make this a draft PR for now? And you can ping us on Slack when it's ready for another look?

deitch:
Yes, I will change it. It originally was draft, I thought we had it ready, now that we have reached the need for more work, I will change it back.

deitch:
Those todos are done. This is ready to go in. It could use future iterations - like finer-grained options on what we report from the graph in various scenarios - but no reason for future improvements to hold us up.

kaniini (approved):

PR #156, opened 2023-04-12
wolfi check update: lets skip validating a pipeline if update config …

rawlingsj:
…is manual

this is because manual updates usually require manual steps in configuring the pipeline fetch or checkout


rawlingsj:
recreating a new PR with more improvements

PR #158, opened 2023-04-12
wolfi check update: ensure update config is provided

rawlingsj:
Now that wolfi has package update config for all packages, let's keep it that way so wolfi is easier to maintain.  It will mean that this is the last time we will need to spend a large amount of time adding update configs to a lot of packages.

### DIFF
--- a/pkg/checks/update.go
+++ b/pkg/checks/update.go
@@ -7,6 +7,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/dprotaso/go-yit"
+
 	version "github.com/wolfi-dev/wolfictl/pkg/versions"
 
 	"github.com/fatih/color"
@@ -44,7 +46,11 @@ func SetupUpdate() (*update.Options, lint.EvalRuleErrors) {
 func (o CheckUpdateOptions) CheckUpdates(files []string) error {
 	updateOpts, checkErrors := SetupUpdate()
 
-	latestVersions, err := updateOpts.GetLatestVersions(o.Dir, GetPackagesToUpdate(files))
+	changedPackages := GetPackagesToUpdate(files)
+
+	validateUpdateConfig(changedPackages, &checkErrors)
+
+	latestVersions, err := updateOpts.GetLatestVersions(o.Dir, changedPackages)
 	if err != nil {
 		addCheckError(&checkErrors, err)
 	}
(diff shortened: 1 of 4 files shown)


pkg/checks/update_test.go
```diff
@@ -136,3 +136,35 @@ func createTestRepo(t *testing.T, dir, tag string) string {
 
 	return c.String()
 }
+
+func TestUpdateKeyExists(t *testing.T) {
```
luhring, line 140: nit: might be nice to separate these test two test cases into separate `t.Run("test case name", ...)` instances to make it easier to review the test output

pkg/checks/update.go
```diff
@@ -64,6 +70,66 @@ func (o CheckUpdateOptions) CheckUpdates(files []string) error {
...
+		}
+
+		if node.Content == nil {
+			addCheckError(checkErrors, fmt.Errorf("config %s has no yaml content", file))
+			continue
+		}
+		// loop over content to ensure an update key exists
+		err = containsKey(node.Content[0], "update")
+		if err != nil {
+			addCheckError(checkErrors, fmt.Errorf("config %s does not have update config provided, see examples in this repository.  Or use update.enabled=false, be aware maintainers may require enabled=true so the package does not become stale", file))
```
luhring, line 101: 👨‍🍳 💋

pkg/checks/update.go
```diff
@@ -64,6 +70,66 @@ func (o CheckUpdateOptions) CheckUpdates(files []string) error {
...
 	return checkErrors.WrapErrors()
 }
+
+// validates update configuration
+func validateUpdateConfig(files []string, checkErrors *lint.EvalRuleErrors) {
+	for _, file := range files {
+		// first need to read raw bytes as unmarshalling a struct without a pointer means update will never be nil
+		if !strings.HasSuffix(file, ".yaml") {
+			file += ".yaml"
+		}
```
luhring, line 80: Just curious why we would need to account for this case (or not be sure about the input state)

luhring (approved):
Nice!

luhring (approved):
Nice!

pkg/checks/update.go
rawlingsj: mainly ease of use, so when running the check command you don't have to keep adding `.yaml`.  With the developer workflow I tend to copy the melange package name and then paste into my terminal to run the command.  Just means folks can specify package name or file name.

PR #167, opened 2023-04-17
wolfi check diff: adds a command to run locally or in CI that checks a newly built apk with the latest in an apk repo

rawlingsj:
This will generate a file containing the diff for packages that can be used to comment on the pull request to make it easier for maintainers to review changes

A PR comment will look like this...
Screenshot 2023-04-17 at 12 54 54

future work: we could modify the check to comment directly on the PR as we have the logic but let's see how this goes first and use an existing action to do the comment.


pkg/checks/diff.go
```diff
@@ -0,0 +1,255 @@
...
+
+func diffDirectories(dir1, dir2 string, newPackages map[string]NewApkPackage) (diffResult, error) {
+	result := diffResult{}
+
+	err := filepath.Walk(dir1, func(path1 string, info1 os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if !info1.IsDir() {
```
luhring, line 129: nit: invert this to remove extra indentation

pkg/checks/diff.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	diffFile := filepath.Join(o.Dir, "diff.log")
+	err = writeDiffLog(rs, diffFile, newPackages)
+	if err != nil {
+		return errors.Wrap(err, "failed writing to file")
+	}
+
+	fmt.Printf("diff written to %s\n", diffFile)
+	return nil
```
luhring, line 104: I think `check ...` commands should probably exit non-zero if a thing being checked is not what's expected. If it's not that kind of command, perhaps it should go somewhere else in the command hierarchy than in `check`.

pkg/cli/check_diff.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+func Diff() *cobra.Command {
+	o := checks.NewDiff()
+	cmd := &cobra.Command{
+		Use:               "diff",
+		DisableAutoGenTag: true,
+		SilenceUsage:      true,
+		SilenceErrors:     true,
+		Short:             "Create a diff comparing proposed apk changes to the latest available in an APKINDEX",
```
luhring, line 18: Is this command expected to be run only after a melange build? That's probably worth calling attention to here 🤔

luhring (changes requested):
Awesome functionality! 👏  Just a couple comments for you

pkg/cli/check_diff.go
rawlingsj: thanks done

pkg/checks/diff.go
rawlingsj: not sure I follow this one, the function does returns errors which fails the check if there's unexpected behaviour.  Unless I'm miss-understanding the comment here?  If I am and it's an easy change mind suggesting an edit so I can follow?

rawlingsj: the errors are also returned here https://github.com/wolfi-dev/wolfictl/pull/167/files#diff-12d09c941117ea670506fe1b328cbbdd29de8f5eb50bab939cde1febeaaf78e4R20

luhring: I might have this wrong, but my understanding was that the `check` commands are kind of like linting, in that they are designed to be helpful as CI checks, and you can rely on them to exit non-zero if the thing being "checked" was not true.

I'm trying to map this understanding onto the `wolfictl check diff` command. What are we checking for? What means the check passed, and what means the check failed? Does the fact that there's a diff mean our expectation wasn't met? This seems to be true for the other `wolfictl check ...` commands: `update` and `soname`.

And if we're not asserting a thing here, I'm wondering if it'd be a better organization of CLI commands (and easier to intuit) to move the `diff` command somewhere else, like maybe directly under `wolfictl`?

rawlingsj: I put it under checks as it shares common code with other checks like soname.  The check diff is more of a utility to help maintainers check differences rather than a failed lint.  Personally I think that's ok but if it's really a blocker I can refactor.

luhring: Not a blocker! Let's move this to the back of our minds as we're thinking about command organization.

PR #171, opened 2023-04-18
wolfi check diff: dont hardcode the architecture when fetching a matching APKINDEX

rawlingsj:


kaniini (approved):

pkg/cli/check_diff.go
```diff
@@ -1,22 +1,36 @@
...
 		RunE: func(cmd *cobra.Command, _ []string) error {
+			arch := ""
+			switch runtime.GOARCH {
+			case "amd64":
+				arch = "x86_64"
+			case "arm64":
+				arch = "aarch64"
+			default:
+				return fmt.Errorf("architecture %s not supported", runtime.GOARCH)
+			}
```
imjasonh, line 31: FWIW apko has `ParseArchitecture` and `ToAPK` to do this for more archs: https://pkg.go.dev/chainguard.dev/apko@v0.7.3/pkg/build/types#Architecture

If it's helpful we could probably do `types.CurrentArchitecture` that looks at `runtime.GOARCH`.

PR #173, opened 2023-04-19
Fix directory descension for configs index

luhring:
(needs to bring in changes from https://github.com/chainguard-dev/melange/pull/395 before it's ready)

This fixes an issue where a configs index was being built to include YAML files from subdirectories when it shouldn't.

This also adds tests around the `NewIndex` function to make sure the right YAML files are getting included or not included.


luhring:
cc: @rawlingsj

rawlingsj (dismissed):

rawlingsj (approved):

imjasonh (approved):

PR #174, opened 2023-04-19
only generate .dot, use graphviz to render it

imjasonh:
Instead of:

```
wolfictl svg
```

...now we need to do:

```
wolfictl dot | dot -Tsvg > out.svg
```

...but this means we have a much lighter (and quieter!) dependency, and can generate PNGs too with `dot -Tpng`


imjasonh:
This will break wolfi-dev/os, which calls `wolfictl svg` here: https://github.com/wolfi-dev/os/blob/main/.github/workflows/ci-build.yaml#L127

We can either remove that entire job, or include `graphviz` in the wolfictl image.

deitch:
> This will break wolfi-dev/os, which calls wolfictl svg here

Can we change os Makefile before this PR to just call `wolfictl dot | dot` now, and only once that is in, merge this?

imjasonh:
> > This will break wolfi-dev/os, which calls wolfictl svg here
> 
> Can we change os Makefile before this PR to just call `wolfictl dot | dot` now, and only once that is in, merge this?

We can't, because `wolfictl dot` doesn't exist yet. We could:

1. add `wolfictl dot` but not remove `wolfictl svg`
2. update os to use `wolfictl dot | dot`
3. remove `wolfictl svg`

Or

1. remove os's call to `wolfictl svg` entirely

deitch:
Ah, I didn't realize. I thought we already had `dot` and `svg`, just removing the other.

I think you have the right approach: open both PRs, when ready, merge this, immediately merge that.

kaniini (approved):

imjasonh:
As an example:

```
go run . dot -d ../wolfi-bootstrap-stage3 | dot -Tpng > out.png
```

produced 
out

PR #180, opened 2023-04-25
wolfictl lint: ensure packages have an epoch set

rawlingsj:
The melange struct for epoch doesn't have an `omitempty` yaml annotation which means it is always 0 when unmarshalling configs.  This lint check validates the raw yaml to make sure epoch is set.

This will help avoid these issues https://github.com/wolfi-dev/os/issues/1569

### DIFF
--- a/pkg/lint/testdata/files/no-epoch.yaml
+++ b/pkg/lint/testdata/files/no-epoch.yaml
@@ -1,0 +1,9 @@
+package:
+  name: no-epoch
+  version: 1.2.3
+  description: "a package that fails as no epoch"
+  copyright:
+    - paths:
+        - "*"
+      attestation: TODO
+      license: GPL-2.0-only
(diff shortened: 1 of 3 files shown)


imjasonh (approved):

pkg/lint/testdata/files/no-epoch.yaml
```diff
@@ -0,0 +1,9 @@
+package:
+  name: no-epoch
+  version: 1.2.3
+  description: "a package that fails as no epoch"
+  copyright:
+    - paths:
```
imjasonh, line 6: You can remove these, which might trigger other lint checks and cause the test to pass erroneously

imjasonh (approved):

pkg/lint/testdata/files/no-epoch.yaml
rawlingsj: Just to confirm which can be removed?  When I remove anything other than the `attestation: TODO` the other lint check fails:
```
valid-copyright-header every package should have a valid copyright header INFO
```

PR #183, opened 2023-04-27
Adds a 'request-version-update' label to all updater PRs and issues

patflynn:


patflynn:
Tested e2e using the CLI manually.

pkg/gh/issues.go
```diff
@@ -114,6 +116,17 @@ func (o GitOptions) CommentIssue(ctx context.Context, owner, repo, comment strin
 	return issue.GetHTMLURL(), err
 }
 
+func (o GitOptions) LabelIssue(ctx context.Context, owner, repo string, number int, labels *[]string) error {
```
luhring, line 119: Is there any sense in testing that the labeling is happening when we expect?

luhring (dismissed):

pkg/update/update.go
```diff
@@ -66,6 +66,7 @@ const (
   <img src="https://raw.githubusercontent.com/wolfi-dev/.github/b535a42419ce0edb3c144c0edcff55a62b8ec1f8/profile/wolfi-logo-light-mode.svg" />
 </p>
 `
+	updateLabel = "request-version-update"
```
kaniini, line 69: Does it make sense to allow the updateLabel to be customized?

pkg/gh/issues.go
patflynn: Are you thinking verifying that the call succeeded here by looking at the returned issue and verifying its state against the arguments, or an end-to-end test to github in the test suite, or both? :)

luhring: Yeah, good point, maybe an E2E test makes more sense. And maybe not now if we don't have a precedent for it. Treat this as optional — LGTM!

pkg/update/update.go
patflynn: https://github.com/wolfi-dev/os/pull/1938

patflynn: Updated the code to pass labels as a flag and tested manually: https://github.com/wolfi-dev/os/pull/1938

rawlingsj (approved):

PR #184, opened 2023-04-27
advisory discover: search NVD directly, instead of with secfixes-tracker

### LINKED ISSUE #96: Discover advisories by querying NVD directly
Today `wolfictl advisory discover` queries the [secfixes service](https://secfixes-tracker-q67u43ydxq-uc.a.run.app/), which periodically queries NVD and our APKINDEX and generates a set of packages in the index that have matching vulns.

Instead of relying on this service, `wolfictl advisory discover` can query these sources and emit findings itself directly. The benefit of this is that we'd effectively rewrite the core of the secfixes service in Go (it's in Python now), and be able to move it around, possibly back up into a service in the future, possibly with notifications.

luhring:
This reimplements the `wolfictl advisory discover` command. The net effect of the command is the same, in that newfound package vulnerabilities are recorded as advisories.

The implementation now **has no dependency on a running instance of the "secfixes-tracker" app**. The command can be run from anywhere that has Internet access.

The performance of this command is noticeably improved when using an NVD API key.

Closes #96 

### Matching logic tradeoffs

As with any matching against unrefined data, a number of tradeoffs are made explicitly in the new matching logic, some biasing toward false positives (FPs) and some toward false negatives (FNs). Also, some of these choices are the same choices made by the secfixes-tracker, while some are different.

I'll summarize these decisions here. Very open to feedback on any of these!

| Idea | Bias | Same as secfixes-tracker? | Implemented? |
|:--- |:---:|:---:|:---:|
| Filter out "sufficiently old" CVEs (e.g. prior to 2015). | FN | DIFFERENT | ✖️ |
| Consider _all_ CPE configurations for a given CVE. | FP | DIFFERENT | ✔️ |
| Ignore CPE configurations with the `AND` operator. | FN | N/A | ✔️ |
| Ignore CPE nodes that are "negated". | FN | DIFFERENT | ✔️ |
| Ignore CPE matches where there's no version data. | FN | ? | ✖️ |
| Filter out CPE responses where `TargetSW` is set but we requested `*`. | FN | N/A | ✔️ |


pkg/advisory/advisory.go
```diff
@@ -168,13 +187,13 @@ func Discover(options DiscoverOptions) error {
 
 			if exists {
 				// TODO: Should we allow for updating existing advisories if previously read vuln
-				// data has been updated?
+				//  data has been updated?
```
imjasonh, line 190: I think we have to somehow, right? If NVD reclassifies a vuln from LOW->CRITICAL, that needs to be reflected.

Maybe it doesn't show up as `vuln.severity=CRITICAL` (and therefore always has been) but instead as a new advisory entry saying that (and when) the severity changed. Our SLO for fixing the critical vuln starts from when it's marked as critical, not from when it was originally detected as low.

(This doesn't have to happen in this PR, just jotting down ideas about it which we can talk about later)

pkg/advisory/advisory.go
```diff
@@ -197,3 +216,44 @@ func Discover(options DiscoverOptions) error {
 
 	return nil
 }
+
+func determinePackagesToLookup(apkindex *repository.ApkIndex, cfgs *configs.Index) []string {
+	packagesFound := make(map[string]struct{})
+	for _, pkg := range apkindex.Packages {
+		if pkg.Origin == "" {
+			// TODO: find out why this case exists in the real APKINDEX
```
imjasonh, line 224: I think this was a melange bug we had a while back. Are recent-ish apks showing up this way?

pkg/cli/advisory_discover.go
```diff
@@ -1,40 +1,51 @@
...
-				http.DefaultClient,
-			)
+			apiKey := p.resolveNVDAPIKey()
 
 			err = advisory.Discover(advisory.DiscoverOptions{
-				Index:                 index,
-				VulnerabilitySearcher: secfixesTracker,
+				Configs:               index,
+				PackageRepositoryURL:  p.packageRepositoryURL,
+				Arch:                  "x86_64",
```
imjasonh, line 39: Are there any weird implications of hard-coding this here? If a package ever exists that's only built for arm64 (unlikely, sure) will it never have vulns discovered?

pkg/vuln/nvdapi/detector.go
```diff
@@ -0,0 +1,434 @@
+//nolint:prealloc // This rule is terrible
```
imjasonh, line 1: I agree. Can we disable it repo-wide?

pkg/vuln/nvdapi/detector.go
```diff
@@ -0,0 +1,434 @@
...
+			for _, cpeMatch := range node.CpeMatch {
+				doCPEsMatch, err := cpeStringsMatch(requestCPE, cpeMatch.Criteria)
+				if err != nil {
+					return nil, err
+				}
+				if !doCPEsMatch {
+					continue
+				}
+
+				if !cpeMatch.Vulnerable {
```
imjasonh, line 177: If you move this above `cpeStringsMatch` you can avoid even checking. Unless you need to first, and I don't understand why.

pkg/vuln/nvdapi/detector.go
```diff
@@ -0,0 +1,434 @@
...
+					continue
+				}
+
+				// TODO: Should we discard CpeMatches with no version data whatsoever?
```
imjasonh, line 181: What would that mean, _all_ versions of the package are vulnerable? 🤔 

If it's nonsense to not have version data, then yeah probably just discard?

pkg/vuln/nvdapi/detector.go
```diff
@@ -0,0 +1,434 @@
...
+		return nil, err
+	}
+
+	// TODO: Deal with pages (not urgent because the default page size is 2,000
+	//  CVEs, and we're searching for single packages at a time.)
+
+	// TODO: Consider using an in-memory cache of API responses, keyed by the
+	//  requested CPE. This would save time on redundant API requests, like searching
+	//  for '...*:go...' multiple times because we've pruned versions from multiple,
+	//  related packages like 'go-1.18', 'go-1.19', and 'go-1.20'.
```
imjasonh, line 244: ❤️ This will be especially important in the private packages repo where we expect to have a lot of old versions in play at a time

pkg/vuln/nvdapi/detector.go
```diff
@@ -0,0 +1,434 @@
...
+
+		return nil, fmt.Errorf("got unexpected response status %d for request to %q. Headers: %+v", s, reqURL, resp.Header)
+	}
+
+	defer resp.Body.Close()
+
+	dec := json.NewDecoder(resp.Body)
+	var cvesResponse CVEsResponse
+	err = dec.Decode(&cvesResponse)
+	if err != nil {
```
imjasonh, line 283:
```suggestion
	var cvesResponse CVEsResponse
	if err := json.NewDecoder(resp.Body).Decode(&cvesResponse); err != nil {
```

(super nit, but it's a lot less code and I think it scans better)

imjasonh (approved):
Nothing blocking, mostly just comments and questions about some TODOs.

My higher-level feedback is that unless we want things to be used as a library by other external Go code, we might want to move this all into `internal` so it doesn't grow unwanted dependents, at least without asking permission first. (That's not blocking either)

Thank you for adding tests for this too! 🙏

pkg/advisory/advisory.go
luhring: Yeah, this is a great point. 💯  I've made a note of this in a scratchpad doc I'll share with you.

I have a lot of thoughts about how to evolve this "event" data structure to account for situations like this, and I want to get advisory data moved into its own repo, and the structs out of Melange, first.

pkg/advisory/advisory.go
luhring: No, I haven't seen it in recent apks. And that makes sense! I'll remove the comment.

pkg/vuln/nvdapi/detector.go
luhring: Yes!!! 🙌

pkg/vuln/nvdapi/detector.go
luhring: I'm not 100% sure I follow — avoid even checking which thing?

pkg/vuln/nvdapi/detector.go
luhring: ❤️

pkg/vuln/nvdapi/detector.go
luhring: 💯

pkg/cli/advisory_discover.go
luhring: No yeah, this is a good callout. Ideally the `discover` command would consult apkindexes for all of our supported architectures, and then use the union set in its logic here.

Is there an authoritative place or way to determine the current set of apkindexes that exist for the distro at a given time?

pkg/vuln/nvdapi/detector.go
```diff
@@ -0,0 +1,434 @@
...
+				if err != nil {
+					return nil, err
+				}
+				if !doCPEsMatch {
+					continue
+				}
+
+				if !cpeMatch.Vulnerable {
+					continue
+				}
```
imjasonh, line 179:
```suggestion
				if !cpeMatch.Vulnerable {
					continue
				}
				
				doCPEsMatch, err := cpeStringsMatch(requestCPE, cpeMatch.Criteria)
				if err != nil {
					return nil, err
				}
				if !doCPEsMatch {
					continue
				}
```

Then if `!cpeMatch.Vulnerable` you never even have to do `cpeStringsMatch`. It's a tiny thing, if you think it's easier to read the other way that's fine, it just bails as quickly and early-ily as possible.

pkg/cli/advisory_discover.go
imjasonh: I don't believe there is. cc @kaniini 

We can invent one, like `archs.json` or something.

pkg/vuln/nvdapi/detector.go
luhring: >What would that mean, all versions of the package are vulnerable? 🤔

Yes, exactly. And I'm _mostly_ seeing this in CVEs from many years ago. Recent CVEs usually do have version data, but not always.

I'll make this adjustment 👍 and we can always revisit if we want to.

pkg/vuln/nvdapi/detector.go
luhring: Ah I got it! I like it, I'll do this. I had thought you meant rearranging the lines would mean some lines weren't needed.

pkg/cli/advisory_discover.go
luhring: Okay cool. For now, I'll change this to be a slice of strings instead of a string, and pass in both `x86_64` and `aarch64` — so that we don't miss anything in the meantime. And then, if it's good with you, we can update to refer to that new authority in the future when it's available.

imjasonh (approved):

PR #193, opened 2023-04-29
New advisory types

luhring:
This PR is intended to:

1. Draft the first iteration of the new approach to the advisory data structures
2. Start using the new data types in existing advisory operations, especially the new "detection event" in the "discover" operation
3. Split out advisory types/functions into separate files


luhring:
Implemented in https://github.com/wolfi-dev/wolfictl/pull/362!

PR #194, opened 2023-05-01
use our own secrets-store CSI driver images

imjasonh:
This moves us toward using our own CSI driver images ([this](https://github.com/chainguard-images/images/tree/main/images/secrets-store-csi-driver) and [this](https://github.com/chainguard-images/images/tree/main/images/secrets-store-csi-driver-provider-gcp)) in our arm64 build cluster.

This PR also bumps the versions of these providers we use to the latest currently provided by Wolfi.

**This introduces a possible circular dependency!** If there is no arm64 image available, we can't build one. The solution would be to remove the `kubectl set image`s and use the upstream image to build the new downstream Chainguard Image, then use that. During normal operation though, since we pin to a specific version, we should be able to build new versions and upgrade manually when needed. However...

**This currently requires manual updates!** If this becomes toilsome or means we miss important releases, we might want to consider automating this with something like digestabot. Oh yeah, this reminds me...

**This currently relies on images by tag!** (the k8s CSI driver image was previously used by tag too, the GCP one by digest). We can pin to an image by digest once one is available, but we should definitely have a digestabot involved to pick up updates in that case.


jdolitsky (approved):
drive it home

jdolitsky (approved):

PR #201, opened 2023-05-06
Add a new lint rule that allows us to force comments to be checked du…

dlorenc:
…ring bumps.

We often have comments like "when bumping, check if CVE mitigations can be removed", but we don't have any way to check that this actually happens. As we further automate package bumps, this is more and more likely to cause us to miss things.

This introduces a simple check that lets maintainers define comments with *hardcoded* version strings. When automation bumps the package.version variable, these will get out of data and the lint will fail.


imjasonh (approved):
Nice

kaniini (approved):

PR #208, opened 2023-05-13
advisory commands: add distro auto-detection, and add flags to disable this detection and interactive prompts

luhring:
## Context

Now that there are multiple distros managed using `wolfictl`, and now that there are two repos per distro (one for the distro itself, with package configs, and one for advisory data), things are getting slightly more complicated. 😰 

Advisory-related commands need to know about the advisories repo (for reading and writing data), and some of these commands also need to know about the distro repo that corresponds to that advisories repo (for reading data). Since multiple local directories are involved, it's not always possible to assume a given operation is intended for the current working directory, because that directory can't be both _the distro repo_ and _the advisories repo_ at the same time.

As of a recent update, users can tell `wolfictl` explicitly where their local clones of the advisories repo and the distro repo are using CLI flags or environment variables. But neither flags nor env vars are always a good fit for running these commands manually. Flags mean extra typing and verbosity for the command, and env vars make it more difficult to switch between distros and easier to make a change in the wrong distro.

## What's been added

This PR attempts to make the user experience less burdensome by adding a distro "auto-detection" mechanism to the advisory commands.  🪄 

The approach here is to see if the current working directory is a distro repo or an advisories repo. If it's neither, the detection fails. Otherwise, it tries to figure out where the other corresponding repo is for the given distro (i.e., where the advisories repo is, if we're currently in the distro repo, or vice versa).

There are also two new flags for these commands:

- `--no-distro-detection`: Don't attempt auto-detection. Fail if any needed repo dir paths haven't been provided by the user (via CLI flags or env vars).
- `--no-prompt`: This doesn't relate to auto-detection per se, but follows suit with ensuring no implicit behavior occurs. This disables interactive user prompts, which is handy for running commands in automated workflows.

Here's a miniature demo of how the distro detection behaves in various scenarios:

image

### DIFF
--- a/pkg/cli/advisory.go
+++ b/pkg/cli/advisory.go
@@ -6,12 +6,14 @@ import (
 	"os"
 	"time"
 
+	"github.com/charmbracelet/lipgloss"
 	"github.com/openvex/go-vex/pkg/vex"
 	"github.com/spf13/cobra"
 	"github.com/wolfi-dev/wolfictl/pkg/advisory"
 	"github.com/wolfi-dev/wolfictl/pkg/advisory/sync"
 	"github.com/wolfi-dev/wolfictl/pkg/configs"
 	advisoryconfigs "github.com/wolfi-dev/wolfictl/pkg/configs/advisory"
+	"github.com/wolfi-dev/wolfictl/pkg/distro"
 )
 
 const (
@@ -56,6 +58,12 @@ func resolveAdvisoriesDir(cliFlagValue string) string {
 	}
 
 	return ""
+}
+
+var distroMutedStyle = lipgloss.NewStyle().Foreground(lipgloss.Color("#888"))
+
+func renderDetectedDistro(d distro.DetectedDistro) string {
+	return distroMutedStyle.Render("Auto-detected distro: ") + d.Name + "\n\n"
 }
 
 func resolveTimestamp(ts string) (time.Time, error) {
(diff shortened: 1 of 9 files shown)


imjasonh (approved):

jspeed-meyers:
Uninformed nerd from the back of the room: Why is `FATA[0000]` the beginning of the error message rather than `FATAL`? Intentional? Inside joke? Silly unicode error?

luhring:
> Uninformed nerd from the back of the room: Why is `FATA[0000]` the beginning of the error message rather than `FATAL`? Intentional? Inside joke? Silly unicode error?

Haha it's not an inside joke, and it's nothing I did in this PR. It's pretty ugly, isn't it.... It looks like logrus's UI to me but let me dig in.

luhring:
> > Uninformed nerd from the back of the room: Why is `FATA[0000]` the beginning of the error message rather than `FATAL`? Intentional? Inside joke? Silly unicode error?
> 
> Haha it's not an inside joke, and it's nothing I did in this PR. It's pretty ugly, isn't it.... It looks like logrus's UI to me but let me dig in.

Looks like that was introduced here: https://github.com/wolfi-dev/wolfictl/commit/226631b9d0f6c52026325f7932f2635b03d4f94f#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261

I don't think this is a crisis, but let me keep this in mind. Using this particular logger is probably unnecessary, since it only ever gets to log the single event when the command returns with an error, and the subcommands usually handle their own error messages just fine.

PR #213, opened 2023-05-16
Bump version of yam

luhring:
Especially to pick up the change from https://github.com/chainguard-dev/yam/pull/16


developer-guy (approved):
LGTM

PR #235, opened 2023-05-29
fix resolve cycle so it can handle multiple paths

### LINKED ISSUE #234: DAG resolveCycle is incomplete and can fail
The current method to resolve a cycle, found [here](https://github.com/wolfi-dev/wolfictl/blob/50669d327980f360f12e395a6a7a2f6b9b6d57e3/pkg/dag/graph.go#L264), uses the following logic:

1. I found a cycle when adding `A -> B`
2. Find the shortest path from `B -> A` (i.e. shortest path that is causing the cycle)
3. Remove the last link (edge) in the shortest path
4. Add `A -> B`, which now should work
5. Add back that removed edge, which now should see that it cannot do what it did before, and resolve to a different dependency

The above works kind of well, but only when two things are true:

* There is only one dependency `B -> A`; if there are multiple, you are stuck
* There is a different way to resolve that last removed link, normally something in upstream; if not, you are stuck

The `resolveCycle()` thus is making some assumptions about how to resolve the dependencies that are not necessarily true.

The current algorithm is improved over a previous one, but still not generic enough. We need a more generic one.

For a practical example, see the below diagram (source graphviz lower down). We want to add `patch:2.7.6-r5@local -> autoconf`. Unfortunately, `autoconf` exists only locally, not upstream in bootstrap, so the only thing it can depend upon is `autoconf;2.71-r2@local`. Except that `autoconf;2.71-r2@local` already depends on `patch:2.7.6-r5@local`, not directly, but via multiple paths.

When it tries to resolve, it finds the shortest path of `autoconf:2.71-r2@local -> busybox:1.36.1-r0@local -> patch:2.7.6-r5@local `. It removes the `busybox -> patch` dependency, tries again... and fails, because another one exists `binutils -> patch`. 

This does not have to be limited to 1 or 2; it could be dozens.

We need a better generic algorithm.

graph

```graphviz
digraph dependency_tree {
    node [shape=box, fontcolor=black];
    "autoconf:2.71-r2@local" [color=red, fontcolor=red];
    "binutils:2.40-r3@local" [color=red, fontcolor=red];
    "patch:2.7.6-r5@local" [color=red, fontcolor=red];
    "build-base:1-r5@local";
    "busybox:1.36.1-r0@local" [color=red, fontcolor=red];
    "ca-certificates-bundle:20230506-r0@local";
    "m4:1.4.19-r4@local" [color=red, fontcolor=red];
    "perl:5.36.1-r0@local";
    "scanelf:1.3.7-r0@local";
    "make:4.3-r3@local";
    "wget:1.21.4-r0@local";
    "binutils:2.39-r1@https://packages.wolfi.dev/bootstrap/stage3/x86_64";

    "autoconf:2.71-r2@local" -> "binutils:2.40-r3@local" [color=red];
    "binutils:2.40-r3@local" -> "patch:2.7.6-r5@local" [color=red];
    "m4:1.4.19-r4@local" -> "binutils:2.40-r3@local" [color=red];
    "busybox:1.36.1-r0@local" -> "patch:2.7.6-r5@local" [color=red];
    "autoconf:2.71-r2@local" -> "build-base:1-r5@local";
    "autoconf:2.71-r2@local" -> "busybox:1.36.1-r0@local" [color=red];
    "autoconf:2.71-r2@local" -> "ca-certificates-bundle:20230506-r0@local";
    "autoconf:2.71-r2@local" -> "m4:1.4.19-r4@local" [color=red];
    "autoconf:2.71-r2@local" -> "perl:5.36.1-r0@local";
    "autoconf:2.71-r2@local" -> "scanelf:1.3.7-r0@local";
    "autoconf:2.71-r2@local" -> "make:4.3-r3@local";
    "autoconf:2.71-r2@local" -> "wget:1.21.4-r0@local";
    "build-base:1-r5@local" -> "busybox:1.36.1-r0@local";
    "build-base:1-r5@local" -> "binutils:2.39-r1@https://packages.wolfi.dev/bootstrap/stage3/x86_64";
    "make:4.3-r3@local" -> "busybox:1.36.1-r0@local";
    "make:4.3-r3@local" -> "binutils:2.39-r1@https://packages.wolfi.dev/bootstrap/stage3/x86_64";
    "busybox:1.36.1-r0@local" -> "binutils:2.39-r1@https://packages.wolfi.dev/bootstrap/stage3/x86_64";
}
```

deitch:
Resolves #234 

`resolveCycle` could only handle a single path. E.g. if we tried to add `A -> B`, and `B` already depended on `A` (direct or indirect) with just one link, then it worked. But if there were multiple paths, it didn't. E.g.:

```
A ---------> B
    --> C ---> B
    --> F ---> B
```

Here A depends on B, but also on C which depends on B, and F which depends on B.

This loops through them until it can resolve them all.

### DIFF
--- a/pkg/dag/graph.go
+++ b/pkg/dag/graph.go
@@ -262,43 +262,58 @@ func (g *Graph) addAppropriatePackage(resolver *apko.PkgResolver, c Package, dep
 // It discovers what the current dependency is that is causing the potential loop,
 // removes the last edge in that cycle, and regenerates that dependency without the previous target.
 func (g *Graph) resolveCycle(c *cycle, dep string, resolver *apko.PkgResolver, localRepoSource string) error {
-	sp, err := graph.ShortestPath(g.Graph, c.target, c.src)
-	if err != nil {
-		return fmt.Errorf("unable to find shortest path: %w", err)
-	}
-	if len(sp) < 2 {
-		return fmt.Errorf("there is no path from %s to %s", c.target, c.src)
-	}
-	// the last edge in the cycle is the one that caused the cycle
-	removeSrc, removeTarget := sp[len(sp)-2], sp[len(sp)-1]
+	// cycle through, removing all "shortest path" until we can resolve the cycle,
+	// then add them back
+	var (
+		removed []cycle
+		origs   []string
+	)
+	for {
+		sp, err := graph.ShortestPath(g.Graph, c.target, c.src)
+		if err != nil {
+			return fmt.Errorf("unable to find shortest path: %w", err)
+		}
+		if len(sp) < 2 {
+			return fmt.Errorf("there is no path from %s to %s", c.target, c.src)
+		}
+		// the last edge in the cycle is the one that caused the cycle
+		removeSrc, removeTarget := sp[len(sp)-2], sp[len(sp)-1]
+		removed = append(removed, cycle{
+			src:    removeSrc,
+			target: removeTarget,
+		})
 
-	edge, err := g.Graph.Edge(removeSrc, removeTarget)
-	if err != nil {
-		return fmt.Errorf("unable to find last edge %s -> %s: %w", removeSrc, removeTarget, err)
+		edge, err := g.Graph.Edge(removeSrc, removeTarget)
+		if err != nil {
+			return fmt.Errorf("unable to find last edge %s -> %s: %w", removeSrc, removeTarget, err)
+		}
+		if edge.Properties.Attributes == nil {
+			return fmt.Errorf("original edge %s -> %s has no attributes", removeSrc, removeTarget)
+		}
+		origs = append(origs, edge.Properties.Attributes["target-origin"])
+		// try to reverse the direction of the edge
+		if err := g.Graph.RemoveEdge(removeSrc, removeTarget); err != nil {
+			return fmt.Errorf("unable to remove original edge %s -> %s: %w", removeSrc, removeTarget, err)
+		}
+		// add in our new edge
+		if err := g.Graph.AddEdge(c.src, c.target, graph.EdgeAttribute("target-origin", dep)); err == nil {
+			break
+		}
 	}
-	if edge.Properties.Attributes == nil {
-		return fmt.Errorf("original edge %s -> %s has no attributes", removeSrc, removeTarget)
-	}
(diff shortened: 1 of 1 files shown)


pkg/dag/graph.go
```diff
@@ -262,43 +262,58 @@ func (g *Graph) addAppropriatePackage(resolver *apko.PkgResolver, c Package, dep
 // It discovers what the current dependency is that is causing the potential loop,
 // removes the last edge in that cycle, and regenerates that dependency without the previous target.
 func (g *Graph) resolveCycle(c *cycle, dep string, resolver *apko.PkgResolver, localRepoSource string) error {
```
luhring, line 264: Would it be feasible to add a test for the expectation for how cycle resolution works?

My impression is that the ability to resolve cycles like this is important, and I'm nervous about myself not being able to absorb all the logical context here fast enough before I submit a change that breaks something. Having a test would help "future me", both in that it'd give examples of what was expected to happen, and in that it'd ensure I didn't break an important aspect here due to my own lack of competence in this area!

deitch: It is a good idea to add them. I usually am strongly in favour. At this point, to be honest, I struggled with how to create a good test for this (which just testifies to the complexity). I spent quite some time collecting the fail case and creating the attached issue just to begin to understand what was happening.

We can try, though

luhring: Okay! I can approve for now (I paused just because I know auto-merge can hide comments sometimes).

But just something to think about, and noting my current level of nervousness (maybe a 5 out of 10 😅 )

luhring (approved):

pkg/dag/graph.go
deitch: No, I would like to actually tackle it. it is important. It cannot really be unit-tested, though, as it involves so much more than just `resolveCycle()`. It cleans some things up and then redoes calls.

We might be able to find a way to test parts of it, and then maybe integration-test the whole thing.

luhring: That makes sense, I figured this might be a higher-level, more-context kind of test

deitch: I think I might have figured out a decent test for this.

deitch: Have a look at #236 @luhring

PR #236, opened 2023-05-29
add cycle resolution test with explanation

deitch:
Does not change the code, but adds a good test for the complex cycle resolution

cc @luhring


pkg/dag/testdata/cycle/README.md
luhring, line 1: AMAZING!

luhring (approved):
Very nice work, thanks @deitch!!!

PR #240, opened 2023-06-01
Add initpkg tool to initialize a new package

kaniini:


pkg/initpkg/context.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	log.Printf("unpacked sources to %s", ctx.WorkDir)
+
+	return nil
+}
+
+// InterpretURI interprets the provided source code URI and fills in project
+// information as well as pipeline instructions in how to fetch the sources.
+func (ctx *Context) InterpretURI(uri string) error {
```
jonjohnsonjr, line 167: I would not use `ctx` as a receiver because it collides with the idiomatic name for`context.Context` parameters.

(I'd also rename the struct from `Context` to something else so we don't have a dozen different flavors of Context floating around, but that's less important than the receiver name.)

pkg/initpkg/context.go
```diff
@@ -0,0 +1,277 @@
...
+	"gopkg.in/yaml.v3"
+)
+
+type Context struct {
+	URI     string
+	Name    string
+	Version string
+	License string
+	Layout  string
+	WorkDir string
```
jonjohnsonjr, line 26: Consider having `Context` taking a `WorkFS` parameter so it's more easily tested with an `apkfs.FullFS` or something instead of always writing to disk.

luhring (commented):
Nice! 🎸 

Docs on how to use w/ example(s) would be awesome! 🙏

Dentrax:
Hey @kaniini! I'm looking forward to get this PR merged. So we can iterate the features over time. Is there anything we should do?

PR #242, opened 2023-06-02
remove 'kubectl set image' from setup-cluster.sh

imjasonh:
Blocked on https://github.com/wolfi-dev/os/pull/2514


jonjohnsonjr (approved):

imjasonh:
The dogfooding config in the downstream repos is now done, and they fail if they have nothing to do, which they don't since they match this previous config.

PR #255, opened 2023-06-09
Add runtime deps to graph

deitch:
The DAG calculation before did not account for runtime dependencies, i.e. those listed in `package.dependencies.runtime` as correctly pointed out by @kaniini. This updates it to account for those. The `.runtime` dependencies are to be resolved solely from the local package list, i.e. no upstream.

In order to make this work, we had to rearrange the package additions to the graph. It makes it easier to understand anyways, which is a good thing. It now is:

1. Do a loop through all packages, adding each package as a graph node, adding each subpackage as a graph node, dependent on the parent
2. Do a loop through all packages, adding each runtime dependency; this only works after the previous step, as all packages and subpackages need to exist as nodes before we can add the runtime dependencies
3. Do a loop through all packages, resolving all environment (build-time) dependencies. This includes local and, if listed, other (upstream) repositories

Steps 1 & 3 already existed, although in a single loop; step 2 is the addition.

Note that this works for everything in `wolfi-dev/os` _except_ for the following errors:

```
ruby3.2-fluentd14-1.14.6-r4: unable to resolve dependency ruby3.2-webrick<1.8: could not find package ruby3.2-webrick<1.8 in indexes
ruby3.2-fluentd15-1.16.1-r1: unable to resolve dependency ruby3.2-webrick<1.8: could not find package ruby3.2-webrick<1.8 in indexes
```

I believe this to be the graph behaving correctly, and the problem is in the package definitions in `wolfi-dev/os`. However, until @kaniini confirms it, this can be approved but should not be merged in. With @kaniini 's confirmation it can be merged in, although if the response is that this algorithm is incorrect, I will fix it.

### DIFF
--- a/pkg/dag/graph.go
+++ b/pkg/dag/graph.go
@@ -74,10 +74,14 @@ func NewGraph(pkgs *Packages, options ...GraphOptions) (*Graph, error) {
 	var arch = "x86_64"
 	localRepo := pkgs.Repository(arch)
 	localRepoSource := localRepo.Source()
+	localOnlyResolver := apk.NewPkgResolver([]apk.NamedIndex{localRepo})
+	resolvers[localRepoSource] = localOnlyResolver
 
-	// 1. go through each known origin package, add it as a vertex
-	// 2. go through each of its subpackages, add them as vertices, with the sub dependent on the origin
-	// 3. go through each of its dependencies, add them as vertices, with the origin dependent on the dependency
+	// the order of adding packages is quite important:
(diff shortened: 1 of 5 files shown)


deitch:
Update: @kaniini confirmed that the devs were wrong, fixed in [wolfi-dev/os PR](https://github.com/wolfi-dev/os/pull/2770).

Once that is in, and this is in, we can start to use the graph in `wolfi-dev/os`

PR #285, opened 2023-07-03
pass resolver along with edges

deitch:
This resolves an issue wherein we would resolve a cycle by using a higher-level resolver, rather than the original one that was used for it.

With this in place, the graph actually works!


kaniini (approved):

deitch:
OK, think I got it, CI and lint should be clean now.

kaniini (approved):

PR #300, opened 2023-07-19
wolfi update: handle upstream projects that recreate a release using …

rawlingsj:
…a tag with a new commit


luhring (approved):
This is excellent, and I love the thoughtful testing approach 💯

PR #301, opened 2023-07-20
add option to load a local grype db

cpanato:


cpanato:
Tested locally and worked!
if you pass a local db it will load and use that one, and in case of a new run and you don't set the parameter it will check if needs to update and in case of a new db is available it will update it.

pkg/cli/scan.go
```diff
@@ -29,13 +29,12 @@ func Scan() *cobra.Command {
 					return fmt.Errorf("failed to open apk file: %w", err)
 				}
 
-				fmt.Println(path.Base(apkFilePath))
+				fmt.Printf("Will process: %s\n", path.Base(apkFilePath))
```
luhring, line 32: Can we make sure this gets written to stderr instead of stdout?

luhring: (we'll want to make sure only the scan output goes to stdout, esp when we support a machine-readable output)

pkg/cli/scan.go
```diff
@@ -60,10 +59,12 @@ func Scan() *cobra.Command {
 
 type scanParams struct {
 	requireZeroFindings bool
+	localDBFile         string
 }
 
 func (p *scanParams) addFlagsTo(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&p.requireZeroFindings, "require-zero", false, "exit 1 if any vulnerabilities are found")
+	cmd.Flags().StringVar(&p.localDBFile, "import-local-grype-db", "", "import a local grype db")
```
luhring, line 67: nit: I'd try to express in the flag name or even just the description text that what we're asking for here is a local filesystem path to the Grype DB

pkg/cli/scan.go
```diff
@@ -60,10 +59,12 @@ func Scan() *cobra.Command {
 
 type scanParams struct {
 	requireZeroFindings bool
+	localDBFile         string
```
luhring, line 62: nit:

```suggestion
	localDBFilePath         string
```

pkg/scan/apk.go
```diff
@@ -87,7 +87,22 @@ func APK(f io.Reader) ([]*Finding, error) {
 
 	syftPkgs := packageCollection.Sorted()
 
-	datastore, _, dbCloser, err := grype.LoadVulnerabilityDB(grypeDBConfig, true)
+	updateDB := true
+	if localDBFile != "" {
+		fmt.Printf("Loading local grype DB %s...\n", localDBFile)
```
luhring, line 92: same comment re: putting progress updates like this onto stderr

luhring (commented):
This is a fantastic addition! 💯  — just a couple minor comments for consideration

pkg/cli/scan.go
cpanato:
```
$ ./wolfictl scan ~/Downloads/os_x86_64_bash-doc-5.2_rc4-r0.apk
Will process: os_x86_64_bash-doc-5.2_rc4-r0.apk
✅ No vulnerabilities found

$ ./wolfictl scan ~/Downloads/os_x86_64_bash-doc-5.2_rc4-r0.apk 2>/dev/null
✅ No vulnerabilities found
```

pkg/scan/apk.go
cpanato: done

luhring (approved):
Amazing 🎉

PR #303, opened 2023-07-21
Add new 'sbom' command

luhring:
Similar to the `wolfictl scan` command, this command lets you target a local APK file and inspect it for packages, using Syft. The default view is a summary outline. But you can also get an SBOM in Syft format using `-o syft-json`. The motivation for supporting this output format is to feed it into a soon-to-come revision of the `scan` command.

Preview:

image

And w/ `-o syft-json`:

image


cpanato (approved):
🚀

PR #312, opened 2023-07-25
wolfi update: fix bug in graphql query where we should be fetching th…

rawlingsj:
…e first set of date ordered tags

### DIFF
--- a/pkg/update/githubReleases.go
+++ b/pkg/update/githubReleases.go
@@ -42,7 +42,7 @@ query {
 {{ range  $index, $r := .RepoList }}
   r{{$r.PackageName}}: repository(owner: "{{$r.Owner}}", name: "{{$r.Name}}") {
     nameWithOwner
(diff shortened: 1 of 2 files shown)


rawlingsj:
verified using https://docs.github.com/en/graphql/overview/explorer
```
# Type queries into this side of the screen, and you will 
# see intelligent typeaheads aware of the current GraphQL type schema, 
# live syntax, and validation errors highlighted within the text.

# We'll get you started with a simple query showing your username!
query { 
  rcbbf241eebedba6fd5e657b3c560fc9dd63c0d3e0de6f816be60e86157ec872c: repository(owner: "apache", name: "kafka") {
    nameWithOwner
    refs(refPrefix: "refs/tags/", query: "", orderBy: {field: TAG_COMMIT_DATE, direction: DESC}, first: 50) {
      totalCount
      nodes {
        name
        target {
          commitUrl
        }
      }
    }
  }
}
```

ajayk (approved):

PR #319, opened 2023-07-31
wolfi update: fix issue where updates come from release monitor so th…

rawlingsj:
…ey have no related commit sha

A recent [change](https://github.com/wolfi-dev/wolfictl/pull/300) to handle upstream projects that re-release using the same tag but different commit sha was added.  However this is causing some PRs to be created incorrectly...

https://github.com/wolfi-dev/os/pull/4042/files
https://github.com/wolfi-dev/os/pull/4039/files
https://github.com/wolfi-dev/os/pull/4043/files

This PR will fix this, if the latest upstream version came from release monitor then there will be no associated commit sha, therefore we should not add to the map of packages to update.

### DIFF
--- a/pkg/update/update.go
+++ b/pkg/update/update.go
@@ -737,6 +737,10 @@ func (o *Options) getPackagesToUpdate(latestVersions map[string]NewVersionResult
 
 		// if versions match but the commit doesn't then we need to update the commit
 		// this can occur when an upstream project recreated a tag with a new commit
+		// if release monitor was used we won't have a commit sha
+		if v.Commit == "" {
+			return results, nil
+		}
 		if currentVersionSemver.Equal(latestVersionSemver) {
 			for i := range pc.Config.Pipeline {
 				pipeline := &pc.Config.Pipeline[i]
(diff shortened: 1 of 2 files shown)


luhring:
>A recent [change](https://github.com/wolfi-dev/wolfictl/pull/300) to handle upstream projects that re-release using the same tag but different commit sha was added. However this is causing some PRs to be created incorrectly...

Just so I'm keeping up, what is it about that change that caused the epoch to reset to `0` in those example PRs?

rawlingsj:
It's because the code previously thought the commit sha was different, as the commit sha is empty when the update comes from release monitor.

When the update comes from GitHub we check to see if the commit sha is different to the one we have in wolfi.  We saw an issue in the last couple of weeks where an upstream project re-released using the same version but a different commit.  

The epoch was incorrectly being set to 0 because the commit sha was different but did not have the bump epoch bool set to true.  

The fix is to not update if there is no new version and there is not a new commit sha.

luhring:
Okay, thanks!

There's something that's still not jibing for me — I think it's the fact that it's ever possible for downstream logic to react to an item in a `map[string]NewVersionResults` by changing a file's epoch from say `2` to `0` without a version bump. It makes sense that this PR's change fixes the symptom, since it avoids creating an item in that map in this scenario, so it doesn't trigger that case, but I'm wondering if there's still a sleeping dragon somewhere... 🤔 

But I think the first priority is fixing the behavior for now, so I'm glad we have this PR!

luhring (approved):

PR #324, opened 2023-08-02
initial commit of `wolfictl build`

imjasonh:
This only works in `--dry-run` mode currently.

Like `wolfictl dag` and `wolfctl text`, this depends on generating a DAG of tasks to execute, but unlike those, it actually "executes" them in order, waiting for dependencies to complete before proceeding through the graph.

Because the graph we generate operates at the _package_ level (incl subpackages) and not the config file level, this produces a graph that's overly fine for our needs; we'd prefer to get a graph of config files to build, which we can add in a subsequent PR.

In any case, this demonstrates the general idea: construct a graph, have each task depend on a channel closing for each dependency, after which point it executes, and after executing, close its channel, signaling dependents, and so on.

This approach doesn't cleanly support limited concurrency, since naively limiting task execution concurrency risks starving the graph for meaningful work (we could randomly pick N blocked tasks and not make progress). In practice if we use this to drive the k8s runner, with ~infinite capacity and supply-side queueing, we shouldn't practically care how quickly we enqueue tasks, and limiting concurrency can be accomplished by limiting available resources.


kaniini (approved):

kaniini:
Looks ok to me but you'll probably want to suppress the lint check.

pkg/cli/build.go
```diff
@@ -0,0 +1,199 @@
...
+	err         error
+	deps        map[string]chan struct{}
+	done, jobch chan struct{}
+}
+
+func (t *task) start(ctx context.Context) {
+	log.Printf("task %q waiting on %q", t.pkg, maps.Keys(t.deps))
+
+	defer close(t.done) // signal that we're done, one way or another.
+	tick := time.NewTicker(30 * time.Second)
```
jonjohnsonjr, line 152:
```suggestion
	tick := time.NewTicker(30 * time.Second)
	defer tick.Stop()
```

pkg/cli/build.go
```diff
@@ -0,0 +1,199 @@
+package cli
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"log"
```
jonjohnsonjr, line 7: I don't have a strong opinion here but elsewhere we use logrus in this repo.

pkg/cli/build.go
```diff
@@ -0,0 +1,199 @@
...
+	// TODO: remove make indirection, invoke melange directly as a library.
+	// TODO: pass --pipeline-dir to melange; until then, skip ko-fips
+	if t.pkg == "ko-fips" {
+		return
+	}
+	c := exec.CommandContext(ctx, "make", "BUILDWORLD=no", "MELANGE_EXTRA_OPTS=--runner=kubernetes", fmt.Sprintf("package/%s", t.pkg)) //nolint:gosec
+	c.Dir = t.dir
+	fmt.Fprintln(t.stderr, c.String())
+	if t.dryrun {
+		time.Sleep(time.Duration(rand.Intn(3000)) * time.Millisecond)
```
jonjohnsonjr, line 183: This seems like it would be useful for testing but needlessly slow down a dryrun.

pkg/cli/build.go
```diff
@@ -0,0 +1,199 @@
...
+			count := len(tasks)
+
+			for _, t := range tasks {
+				if err := t.wait(ctx); err != nil {
+					return fmt.Errorf("failed to build %s: %w", t.pkg, err)
+				}
+				delete(tasks, t.pkg)
+				log.Printf("DONE %s (%d/%d)", t.pkg, count-len(tasks), count)
+			}
+			log.Println("ALL DONE!!")
```
jonjohnsonjr, line 125: All caps is frightening.

pkg/cli/build.go
```diff
@@ -0,0 +1,200 @@
...
+				}
+				delete(tasks, t.pkg)
+				log.Printf("Finished building %s (%d/%d)", t.pkg, count-len(tasks), count)
+			}
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVarP(&dir, "dir", "d", ".", "directory to search for melange configs")
+	cmd.Flags().StringVar(&pipelineDir, "pipeline-dir", "", "directory used to extend defined built-in pipelines")
```
mattmoor, line 131: Given the TODO below should we drop this for now?

pkg/cli/build.go
```diff
@@ -0,0 +1,200 @@
...
+					archs:       archs,
+					dryrun:      dryrun,
+					done:        make(chan struct{}),
+					deps:        map[string]chan struct{}{},
+					jobch:       jobch,
+				}
+			}
+
+			if len(args) == 1 {
+				// Build only this one package.
```
mattmoor, line 52: What if dependencies need to be rebuilt? 🤔

pkg/cli/build.go
```diff
@@ -0,0 +1,200 @@
...
+				return err
+			}
+
+			tasks := map[string]*task{}
+			for _, pkg := range g.Packages() {
+				if tasks[pkg] == nil {
+					tasks[pkg] = newTask(pkg)
+				}
+				for k, v := range m {
+					if strings.HasPrefix(k, pkg+":") {
```
mattmoor, line 96: This could use a comment

pkg/cli/build.go
```diff
@@ -0,0 +1,200 @@
...
+						}
+					}
+				}
+			}
+
+			if len(tasks) == 0 {
+				return fmt.Errorf("no packages to build")
+			}
+
+			// TODO: limit concurrency, without starving the graph.
```
mattmoor, line 113: `errgroup` with `SetLimit`?

pkg/cli/build.go
mattmoor: I see it used in the graph computation, so I guess not.

mattmoor (approved):
disabled automerge for comments, but LGTM

pkg/cli/build.go
imjasonh: That's going to be a followup PR, where I add `--buildworld` (whether or not to use deps from the package repo or build them ourselves) and `--todo-needs-a-name` (whether or not to build dependent packages)

If args are omitted we'll build everything.

pkg/cli/build.go
imjasonh: I'll squeeze this into #353 so I don't undo your approval.

PR #330, opened 2023-08-07
fix: use correct yam config for non-cwd advisories

luhring:
Previously, the use of `.AutomaticConfig()` was reading the yam config from the current working directory, even if `wolfictl` was interacting with advisory data in a _different_ directory, and the net result was that the correct formatting wasn't used for the advisories directory.

cc: @ajayk @kaniini

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/anchore/syft v0.86.1
 	github.com/chainguard-dev/go-apk v0.0.0-20230710230135-7fc46e8b3c4d
 	github.com/chainguard-dev/kontext v0.1.0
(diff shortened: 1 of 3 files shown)


luhring:
🤨 

```
--- FAIL: TestBuildConfigsIndex (0.00s)
    build_test.go:36: 
        	Error Trace:	/home/runner/work/wolfictl/wolfictl/pkg/configs/build/build_test.go:36
        	Error:      	Received unexpected error:
        	            	unable to update "config-2.yaml": unable to process and update index entry for "config-2.yaml": file does not exist
        	Test:       	TestBuildConfigsIndex
FAIL
```

kaniini (approved):

PR #356, opened 2023-08-23
wolfictl lint: ensure packages that have git-checkout use github upda…

rawlingsj:
…te config

Using GitHub update config is required as we need to get the expected-commit from the graphql api


luhring (approved):
Nice!

PR #357, opened 2023-08-23
wolfi updates: add strip preffix and suffix for release monitor updates

rawlingsj:


rawlingsj:
this will mean we can enable automated updates for things like https://github.com/wolfi-dev/os/issues/4662

PR #358, opened 2023-08-23
Fix: segfault when `relatedMetadata` is nil

pdeslaur:
This bug repros with this: `wolfictl scan =(curl -sL https://packages.wolfi.dev/os/x86_64/nodejs-20-20.4.0-r0.apk)`


luhring:
Thanks @pdeslaur!

PR #361, opened 2023-08-29
Add option to export in YAML

pdeslaur:
This change adds the option to export all advisories in a single YAML output:

```
wolfictl adv export --format yaml
```

I haven't added a test for this yet. The output order of the map of events isn't predictable yet, and it causes some instability. The format for advisories V2 will allow to test this a lot easier!

### DIFF
--- a/pkg/advisory/export.go
+++ b/pkg/advisory/export.go
@@ -4,19 +4,21 @@ import (
 	"bytes"
 	"encoding/csv"
 	"io"
+	"log"
 	"sort"
 
 	"github.com/samber/lo"
 	"github.com/wolfi-dev/wolfictl/pkg/configs"
 	"github.com/wolfi-dev/wolfictl/pkg/configs/advisory"
+	"gopkg.in/yaml.v2"
 )
 
 type ExportOptions struct {
 	AdvisoryCfgIndices []*configs.Index[advisory.Document]
 }
 
-// Export returns a reader of advisory data encoded as CSV.
-func Export(opts ExportOptions) (io.Reader, error) {
+// ExportCSV returns a reader of advisory data encoded as CSV.
+func ExportCSV(opts ExportOptions) (io.Reader, error) {
 	buf := new(bytes.Buffer)
 	csvWriter := csv.NewWriter(buf)
 	defer csvWriter.Flush()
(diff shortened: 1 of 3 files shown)


kaniini (approved):

pkg/cli/advisory_export.go
```diff
@@ -67,6 +77,7 @@ func AdvisoryExport() *cobra.Command {
 				defer outputFile.Close()
 			}
 
+			fmt.Println(export)
```
luhring, line 80: What are you thinking for this line? I think we're already outputting the exported data using the user's `p.outputLocation` selection, is that right?

pkg/cli/advisory_export.go
```diff
@@ -86,12 +97,27 @@ type exportParams struct {
...
+
 func (p *exportParams) addFlagsTo(cmd *cobra.Command) {
 	addNoDistroDetectionFlag(&p.doNotDetectDistro, cmd)
 
 	cmd.Flags().StringSliceVarP(&p.advisoriesRepoDirs, "advisories-repo-dir", "a", nil, "directory containing an advisories repository")
 
 	cmd.Flags().StringVarP(&p.outputLocation, "output", "o", "", "output location (default: stdout)")
+
+	cmd.Flags().StringVarP(&p.format, "format", "f", "",
+		fmt.Sprintf("Output format. One of: [%q, %q, %q]", OutputDefault, OutputYAML, OutputCSV))
```
luhring, line 122: nit: I might use Cobra's mechanism for setting the default value, rather than managing it ourselves. This also means we won't suggest `""` as a value for `--format` in the help text, which I think is a nice (but super tiny) improvement in clarity. So something like this:

```suggestion
	cmd.Flags().StringVarP(&p.format, "format", "f", OutputCSV,
		fmt.Sprintf("Output format. One of: [%s]", strings.Join([]string{OutputYAML, OutputCSV}, ", ")))
```

Also, if you cascade this change to lines 59 and 62 as well, you end up not needing the `OutputDefault` value at all

pkg/advisory/export.go
```diff
@@ -4,19 +4,21 @@ import (
 	"bytes"
 	"encoding/csv"
 	"io"
+	"log"
 	"sort"
 
 	"github.com/samber/lo"
 	"github.com/wolfi-dev/wolfictl/pkg/configs"
 	"github.com/wolfi-dev/wolfictl/pkg/configs/advisory"
+	"gopkg.in/yaml.v2"
```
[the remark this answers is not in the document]
pdeslaur: Fixed!

pkg/cli/advisory_export.go
pdeslaur: Whoops, that should have been removed.

pkg/cli/advisory_export.go
```diff
@@ -86,12 +97,27 @@ type exportParams struct {
...
+
+	// format controls how commands will produce their output.
+	format string
 }
 
+const (
+	// OutputDefault default output.
+	OutputDefault = ""
+	// OutputJSON JSON output.
+	OutputYAML = "yaml"
```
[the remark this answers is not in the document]
pdeslaur: Fixed!

pkg/cli/advisory_export.go
pdeslaur: Cleaned this up - great suggestion!

pdeslaur (commented):
Thanks!

luhring (approved):
Nice enhancement!!

PR #362, opened 2023-08-31
Upgrade advisory schema to v2

luhring:
This PR introduces formality to the concept of a "schema" for advisory documents, and declares a new schema, version `2`, that's now used throughout the advisory-related functions of `wolfictl`.

⚠️  **THIS IS A BREAKING CHANGE.** As of this PR, `wolfictl` is no longer able to operate on data using the "v1" (original) data format used in advisory documents.

Commands updated to v2 schema:

- `wolfictl adv create`
- `wolfictl adv db`
- `wolfictl adv discover`
- `wolfictl adv list`
- `wolfictl adv migrate` (new)
- `wolfictl adv validate`
- `wolfictl scan`

Planned follow-ups after this PR:

- Ability to add alias information to advisories
- Documentation around how to perform future schema updates


luhring (commented):
(from live review w/ @pdeslaur)

pkg/configs/advisory/v2/event.go
```diff
@@ -0,0 +1,181 @@
...
+	err := pe.Data.Decode(data)
+	if err != nil {
+		return Event{}, err
+	}
+	event.Data = *data
+
+	return event, nil
+}
+
+func (e Event) Validate(eventIndex int) error {
```
pdeslaur, line 116: I generally don't like having `Validate` needing a parameter to support a style of calling the method. The current implementation assumes that `Validate` would be called in a `loop`, using an index. I ended up here after looking at the `Validate` method on `pkg/advisory/request.go` and wondering why we provided `0`!

So - totally not a big deal, but if we could lift `fmt.Sprintf("(index %d)", eventIndex)` out of this method, the code would be cleaner!

pkg/configs/rwfs/os/memfs/memfs.go
pdeslaur, line 1: Didn't really review this file - but I'm assuming it works :)

Is that written from scratch or taken from somewhere?

pkg/scan/filter.go
```diff
@@ -60,17 +64,21 @@ func FilterWithAdvisories(result *Result, advisoryCfgs *configs.Index[advisoryco
...
-			adv := packageAdvisories[finding.Vulnerability.ID]
-
-			if advisory.IsResolved(adv, result.TargetAPK.Version) {
-				return false
+			adv, ok := packageAdvisories.GetByVulnerabilityID(finding.Vulnerability.ID)
+			if ok {
+				if adv.ResolvedAtVersion(result.TargetAPK.Version) {
+					return false
+				}
 			}
```
pdeslaur, line 72:
```suggestion
			if ok && adv.ResolvedAtVersion(result.TargetAPK.Version) {
				return false
			}
```

pkg/scan/filter.go
```diff
@@ -38,17 +37,22 @@ func FilterWithAdvisories(result *Result, advisoryCfgs *configs.Index[advisoryco
...
-			// If the advisory contains any entries, filter it out!
-			if len(adv) >= 1 {
-				return false
+			adv, ok := packageAdvisories.GetByVulnerabilityID(finding.Vulnerability.ID)
+			if ok {
+				// If the advisory contains any events, filter it out!
+				if len(adv.Events) >= 1 {
+					return false
+				}
 			}
```
pdeslaur, line 46:
```suggestion
			// If the advisory contains any events, filter it out!
			if ok && len(adv.Events) >= 1 {
				return false
			}
```

pkg/scan/filter.go
```diff
@@ -60,17 +64,21 @@ func FilterWithAdvisories(result *Result, advisoryCfgs *configs.Index[advisoryco
 
 	case AdvisoriesSetResolved:
 		resultFindings := lo.Filter(result.Findings, func(finding *Finding, _ int) bool {
-			adv := packageAdvisories[finding.Vulnerability.ID]
-
-			if advisory.IsResolved(adv, result.TargetAPK.Version) {
-				return false
+			adv, ok := packageAdvisories.GetByVulnerabilityID(finding.Vulnerability.ID)
```
pdeslaur, line 67: My 2c: it'd be cool to have a method `GetByAlias` or something like that. It could return a match if the vulnerability ID or an alias matches.

pkg/advisory/export_test.go
```diff
@@ -9,11 +9,60 @@ import (
...
+
+				if diff := cmp.Diff(string(expectedBytes), string(exportedBytes)); diff != "" {
+					t.Errorf("ExportCSV() produced unexpected data (-want +got):\n%s", diff)
+				}
+			}
+		})
+	}
+}
+
+func TestExportYAML(t *testing.T) {
```
pdeslaur, line 65: Do we still need `TestExportYAML`? Looks like it's folded in a really nice way into the method above!

pkg/advisory/request.go
```diff
@@ -26,39 +22,23 @@ func (req Request) Validate() error {
...
-		Status:          req.Status,
-		Justification:   req.Justification,
-		ImpactStatement: req.Impact,
-		ActionStatement: req.Action,
-		FixedVersion:    req.FixedVersion,
+	if req.Event.IsZero() {
+		return errors.New("event cannot be zero")
 	}
+
+	return req.Event.Validate(0)
```
pdeslaur, line 43: I left a comment elsewhere but was confused about the `0` of `Validate(0)`

pkg/cli/components/advisory/prompt/model.go
```diff
@@ -46,14 +46,8 @@ func (m Model) newPackageFieldConfig() field.TextFieldConfiguration {
...
-var cveIDRegex = regexp.MustCompile(`^CVE-\d{4}-\d{4,}$`)
-
-// grabbed this regex from Github Docs
-// see https://docs.github.com/en/code-security/security-advisories/global-security-advisories/about-the-github-advisory-database#about-ghsa-ids
-var ghsaRegex = regexp.MustCompile(`GHSA(-[23456789cfghjmpqrvwx]{4}){3}`)
-
-var ValidCVEID = field.TextValidationRule(func(value string) error {
-	if !cveIDRegex.MatchString(value) && !ghsaRegex.MatchString(value) {
+var validCVEID = field.TextValidationRule(func(value string) error {
+	if !vuln.RegexCVE.MatchString(value) && !vuln.RegexGHSA.MatchString(value) {
```
pdeslaur, line 50: Should we use `vuln.ValidateID`, or do we specifically not want to allow GO entries

pkg/configs/advisory/v2/advisory.go
```diff
@@ -0,0 +1,123 @@
...
+		return Event{}
+	}
+
+	sorted := adv.SortedEvents()
+	return sorted[len(adv.Events)-1]
+}
+
+// SortedEvents returns the events in the advisory, sorted by timestamp, from
+// oldest to newest.
+func (adv Advisory) SortedEvents() []Event {
```
pdeslaur, line 36: I would specifically document that we're sorting in-place - and calling this method will have side effects!

pkg/configs/advisory/v2/error.go
```diff
@@ -0,0 +1,31 @@
+package v2
+
+import "fmt"
+
+type labeledError struct {
+	label string
+	err   error
+}
+
+// Error implements the error interface.
```
pdeslaur, line 10: Is the comment meaning to say `labeledError implements the error interface`?

pkg/configs/rwfs/os/memfs/memfs.go
luhring: Written from scratch by ChatGPT and then massaged by me :D

pkg/configs/advisory/v2/advisory.go
luhring: Unless you object, I'm going to update the method to ensure the slice is copied before sorting happens so there are no side effects!

pkg/advisory/advisory.go
```diff
@@ -4,7 +4,7 @@ import (
 	"sort"
 
 	"github.com/openvex/go-vex/pkg/vex"
-	advisoryconfigs "github.com/wolfi-dev/wolfictl/pkg/configs/advisory"
+	advisoryconfigs "github.com/wolfi-dev/wolfictl/pkg/configs/advisory/v1"
```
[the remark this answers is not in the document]
luhring: This whole file can be deleted now!

pkg/scan/filter.go
luhring: IIUC, that's the intent of this method (renamed to drop the `ID` suffix):

```go
// GetByVulnerability returns the advisory that references the given
// vulnerability ID as its advisory ID or as one of the advisory's aliases. If
// such an advisory does not exist, the second return value will be false;
// otherwise it will be true.
```

pkg/advisory/export_test.go
luhring: Great catch!

pkg/cli/components/advisory/prompt/model.go
luhring: Good question. We don't have good reason to disallow GO- entries, we just haven't used them yet. But we're already using alternative IDs (i.e. GHSA), so we might as well allow GO here, too. I'll update this to use `vuln.ValidateID`.

luhring: oooo this ends up collapsing this stanza down to:

```go
var validCVEID = field.TextValidationRule(vuln.ValidateID)
```

🤩

pkg/configs/advisory/v2/error.go
luhring: Yeah, basically. It's a doc comment, so the first word matches the method name. Adding this method implements the `error` interface on `labeledError`'s behalf. Let me try an update for clarity...

pkg/configs/advisory/v2/advisory.go
pdeslaur: Works for me!

pkg/configs/advisory/v2/error.go
pdeslaur: Ohhh, right. Yeah, that's a standard golang thing - I forgot.

pkg/scan/filter.go
pdeslaur: Oh neat - you were one step ahead of me 😎

pdeslaur (approved):

pdeslaur (dismissed):

pdeslaur (commented):
Such green! I'm good to go. I'll wait until Monday to approve to ensure we don't merge by accident over the weekend!

So pumped about this!

pdeslaur (approved):

pdeslaur (approved):

PR #371, opened 2023-09-07
remove `wolfictl pod`

imjasonh:
Goodnight sweet prince. You will not be missed 🖖


kaniini:
Can we actually do this yet?  AFAIK `enterprise-packages` still uses `wolfictl pod`?

mattmoor:
lol, I came to say that 🤣

PR #375, opened 2023-09-08
Bump version of syft

pdeslaur:


go.mod
```diff
@@ -1,13 +1,13 @@
 module github.com/wolfi-dev/wolfictl
 
-go 1.21
+go 1.21.0
```
luhring, line 3: Why this change? Just curious

pdeslaur: Not sure! I just did `go mod tidy` and it added the patch version 🤔

luhring: Okay! This is the closest I found for now: https://tip.golang.org/doc/go1.21#tools 🤷

PR #376, opened 2023-09-08
fix: minor issues with advisory prompts

luhring:
- should prompt for "note" value for new event types: fix-not-planned, analysis-not-planned
- prompt should allow empty "note" response for true positives, since the field is optional
- prompt should require a non-empty "note" response for false positives
- non-empty response values were followed by the prompt askig for the same field again
- change event index from 0-based to 1-based in advisory validation messages

### DIFF
--- a/pkg/cli/components/advisory/field/field.go
+++ b/pkg/cli/components/advisory/field/field.go
@@ -8,6 +8,7 @@ import (
 )
 
 type Field interface {
+	ID() string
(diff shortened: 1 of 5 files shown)


pdeslaur (approved):
Cool! So this basically allows an empty note?

luhring:
> Cool! So this basically allows an empty note?

For true positives, yes

PR #382, opened 2023-09-11
don't call subgraph methods, they're broken

imjasonh:
`SubgraphWith{Leaves,Roots}` doesn't seem to work, and aren't effectively called by anything today anyway. Remove those calls while Jon fixes the methods, and we can readd them.


luhring (approved):
😢

imjasonh:
> 😢

They'll be back!

PR #385, opened 2023-09-13
wolfictl update: transform-version needs to run before versions are sorted and compared

rawlingsj:


luhring:
One thought: I'm wondering if it makes sense for all users of Melange configs to go through Melange's central parsing function (`ParseConfiguration` IIRC) before accessing data, so we don't have to keep multiple processing implementations in sync (e.g. for var transforms). WDYT?

rawlingsj:
> WDYT?

I think that sounds like a separate PR TBH, this is currently fixing broken functionality.

luhring:
> > WDYT?
> 
> I think that sounds like a separate PR TBH, this is currently fixing broken functionality.

Roger — we'll get this merged then. Maybe we could look at unifying the code paths some time in the future if it still makes sense

luhring (approved):

PR #390, opened 2023-09-14
always use buildtime repos for runtime resolution

imjasonh:
This was previously a flag in `wolfictl text`, but I think it should just always be true.

If this lands well I will remove the option to set it to false and always have it be true.


joshrwolf (approved):

joshrwolf:
lgtm, validated this makes the following wai: https://github.com/chainguard-dev/enterprise-packages/pull/533

PR #409, opened 2023-09-25
bump melange dependency

rawlingsj:


cpanato (approved):
should we also bump in the enterprise-packages?

rawlingsj:
> should we also bump in the enterprise-packages?

once this merges we can release a new image and any repos that use wolfictl

PR #414, opened 2023-10-04
wolfictl lint: allow flag to override what severity of lint rules to …

rawlingsj:
…evaluate

This can be useful if a user wants to create CI jobs with different required checks for different severity levels

This also makes the copywrite rule an error as this should be mandatory


cpanato (approved):

rawlingsj:
not needed

PR #415, opened 2023-10-04
Revert "wolfictl update: transform-version needs to run before versio…

rawlingsj:
…ns are sorted and compared"

This reverts commit bb20ed5595d97624033c9557662e4eeac06cdfa5.

There's more edge cases to consider which will need more test cases.  We've seen an unexpected version update PR come in that uses the version transforms https://github.com/wolfi-dev/os/pull/6255.  Let's roll this commit back and add more tests to cover more edge cases.

### DIFF
--- a/pkg/update/githubReleases.go
+++ b/pkg/update/githubReleases.go
@@ -624,11 +624,6 @@ func (o GitHubReleaseOptions) prepareVersion(nameHash, v, id string) (string, er
 		return "", nil
 	}
 
-	v, err := transformVersion(c.Update, v)
-	if err != nil {
-		return "", errors.Wrapf(err, "failed to transform version %s", v)
-	}
-
 	return v, nil
 }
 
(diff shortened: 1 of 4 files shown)


rawlingsj:
Sigh - need to think about this a bit more and look at more test cases.  The original implemenation of version transforms means we were missing some updates

```
wolfictl check update  --override-version 0.0.1 openjdk-21
```
original `21.9` is wrong it should be `[jdk-21-ga](https://github.com/openjdk/jdk21u/releases/tag/jdk-21-ga)` however I think that's partly down to an issue with the [regex](https://github.com/wolfi-dev/os/blob/f30de34b18e3d17c4102ee8f7e2ac538c4f212a9/openjdk-21.yaml#L226-L228).  If you ignore the regex issue preventing`-ga`  then the fix my change made would work, as it would get the next most recent release [jdk-21+35](https://github.com/openjdk/jdk21u/releases/tag/jdk-21%2B35).

i.e. the code in wolfictl was incorrectly resolving `21.9` which it should be `21.35`


So that seemed to be working but we've had a slightly different case just happed for openjdk-20 (which has now been archived last month https://github.com/openjdk/jdk20u)

With my fix the bot created a PR for an old version, the opposite of what was happening with my fix above.

TL;DR I'm going to familiarise myself with the version transform code that was added + get some more test cases that cover these odd versions we're seeing on openjdk.

Leaving as draft for now until I really understand what's going on.

PR #417, opened 2023-10-06
fix: find advisories for origin package during scan filtering

luhring:
This was something I overlooked in the initial implementation of the filtering flags (i.e. `-f`/`-a`) for `wolfictl scan`.

Now, when scanning a subpackage APK, and applying advisory data to filter out scan results, the command will look up the origin package for the APK, and use that to find advisory entries. (Advisory data is always filed under the origin package name, never for the subpackage directly.)

cc: @vaikas

### DIFF
--- a/pkg/scan/apk.go
+++ b/pkg/scan/apk.go
@@ -49,8 +49,19 @@ type Result struct {
 }
 
 type TargetAPK struct {
-	Name    string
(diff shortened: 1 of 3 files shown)


vaikas (approved):
Awesomesauce! Thanks for the quick fix!

PR #418, opened 2023-10-09
fix: refuse to unmarshal advisories with unknown fields

luhring:
It turns out that there's a [bug](https://github.com/go-yaml/yaml/issues/460) in `yaml.v3` where setting `.KnownFields(true)` will reject unknown fields during unmarshaling **except when custom unmashalers are used**. In fact, there's no first-class way in the library to reject unknown fields while implementing customer unmarshalers.

To overcome this, this PR introduces a new utility function `strictUnmarshal` that can be used during custom unmarshaling to achieve the same effect that `.KnownFields` has on root node decoding.

Thanks to @mamccorm for helping me find this! 🙇 

Now, when a field outside of the v2 schema is used, `wolfictl adv validate` fails with an error. For example:

```console
$ w adv validate
Auto-detected distro: Wolfi

FATA[0000] error during command execution: unable to parse configuration at "keycloak.advisories.yaml": strict YAML unmarshaling failed: failed to decode to Go type *v2.Detection: strict YAML unmarshaling failed: failed to decode to Go type *v2.partialDetection: yaml: unmarshal errors:
  line 2: field note not found in type v2.partialDetection
```


mamccorm (approved):
thanks for fixing this one

PR #419, opened 2023-10-09
feat: add rich support for advisory aliases

luhring:
This PR adds two new commands:

- `wolfictl adv alias discover`: looks up aliases in supported vulnerability data sources (for now, just GHSA) to populate the `aliases` field on existing advisories.
- `wolfictl adv alias find`: a simple utility command that looks up aliases for the vulnerability IDs provided as arguments on the command line. (This command doesn't touch any advisory data.)

This PR makes it easier to ensure we stay aware of relevant vulnerability records in the upstream data sources we care about.

### Future work

This is already a large PR, but future PRs could look at the following enhancements to take this PR's functionality further...

1. Automatically discover aliases as part of the `wolfictl adv create` command.
2. Automatically discover aliases as part of the `wolfictl adv discover` command.
3. Add support for the [Go Vulnerability Database](https://go.dev/security/vuln/database) as a source of aliases.
4. Add support for the [PSF Advisory Database](https://github.com/psf/advisory-database) as a source of aliases.
5. Add a flag to the `wolfictl adv alias discover` command to perform discovery only on advisories that have no alias data yet.
6. Cache API queries in memory during command execution, since there's a decent amount of redundancy today of vulnerability IDs in our advisory data set.

### DIFF
--- a/pkg/advisory/alias_finder.go
+++ b/pkg/advisory/alias_finder.go
@@ -60,11 +60,11 @@ func (f *HTTPAliasFinder) GHSAsForCVE(ctx context.Context, cveID string) ([]stri
 		return nil, err
 	}
 
-	cveIDs := lo.Map(ghsas, func(ghsa githubSecurityAdvisoryResponse, _ int) string {
+	ghsaIDs := lo.Map(ghsas, func(ghsa githubSecurityAdvisoryResponse, _ int) string {
 		return ghsa.GHSAID
 	})
 
-	return cveIDs, nil
+	return ghsaIDs, nil
 }
 
 func (f *HTTPAliasFinder) gitHubAPIGet(ctx context.Context, requestPath string, queryParameters map[string]string) (io.ReadCloser, error) {
--- a/pkg/advisory/alias_finder.go
+++ b/pkg/advisory/alias_finder.go
@@ -1,0 +1,176 @@
+package advisory
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"time"
+
+	"github.com/samber/lo"
+)
+
+type AliasFinder interface {
+	CVEForGHSA(ctx context.Context, ghsaID string) (string, error)
+	GHSAsForCVE(ctx context.Context, cveID string) ([]string, error)
+}
+
+type HTTPAliasFinder struct {
+	client *http.Client
+}
+
+func NewHTTPAliasFinder(client *http.Client) *HTTPAliasFinder {
+	return &HTTPAliasFinder{
+		client: client,
+	}
+}
+
+func (f *HTTPAliasFinder) CVEForGHSA(ctx context.Context, ghsaID string) (string, error) {
+	requestPath := fmt.Sprintf("/advisories/%s", url.PathEscape(ghsaID))
+	respBody, err := f.gitHubAPIGet(ctx, requestPath, nil)
+	if err != nil {
+		return "", err
+	}
+	defer respBody.Close()
+
+	var ghsa githubSecurityAdvisoryResponse
+	if err := json.NewDecoder(respBody).Decode(&ghsa); err != nil {
+		return "", err
+	}
+
+	return ghsa.CVEID, nil
+}
+
+func (f *HTTPAliasFinder) GHSAsForCVE(ctx context.Context, cveID string) ([]string, error) {
+	respBody, err := f.gitHubAPIGet(
+		ctx,
+		"/advisories",
+		map[string]string{"cve_id": cveID},
+	)
+	if err != nil {
+		return nil, err
+	}
+	defer respBody.Close()
+
+	var ghsas []githubSecurityAdvisoryResponse
+	if err := json.NewDecoder(respBody).Decode(&ghsas); err != nil {
+		return nil, err
+	}
+
+	cveIDs := lo.Map(ghsas, func(ghsa githubSecurityAdvisoryResponse, _ int) string {
+		return ghsa.GHSAID
+	})
+
+	return cveIDs, nil
+}
+
+func (f *HTTPAliasFinder) gitHubAPIGet(ctx context.Context, requestPath string, queryParameters map[string]string) (io.ReadCloser, error) {
+	u, err := url.Parse("https://api.github.com")
+	if err != nil {
+		return nil, err
+	}
+	u.Path = requestPath
+	if queryParameters != nil {
+		q := u.Query()
+		for k, v := range queryParameters {
+			q.Set(k, v)
+		}
+		u.RawQuery = q.Encode()
+	}
+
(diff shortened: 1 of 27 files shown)


pkg/advisory/alias_finder.go
```diff
@@ -0,0 +1,176 @@
...
+			//nolint:errcheck // This is a best-effort attempt to read the body.
+			msg, _ = io.ReadAll(resp.Body)
+		}
+		return nil, fmt.Errorf("unexpected status code %d, body: %s", resp.StatusCode, string(msg))
+	}
+
+	return resp.Body, nil
+}
+
+type githubSecurityAdvisoryResponse struct {
```
pdeslaur, line 114: Not really actionable for now, but it would be neat to see most of the logic talking to GitHub incorporated into https://github.com/google/go-github so we can reuse it later. We'll need more integration with GitHub and having a structured way of fetching interacting with GitHub would be helpful long term.

Looks like Billy contributed a bunch to go-github repo

pkg/advisory/alias_finder.go
```diff
@@ -0,0 +1,176 @@
...
+
+	resp, err := f.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusOK {
+		var msg []byte
+		if resp.Body != nil {
+			//nolint:errcheck // This is a best-effort attempt to read the body.
+			msg, _ = io.ReadAll(resp.Body)
```
pdeslaur, line 106: Can we panic if it fails instead of ignoring the error?

pkg/advisory/alias_finder.go
```diff
@@ -0,0 +1,176 @@
...
+		return nil, err
+	}
+	defer respBody.Close()
+
+	var ghsas []githubSecurityAdvisoryResponse
+	if err := json.NewDecoder(respBody).Decode(&ghsas); err != nil {
+		return nil, err
+	}
+
+	cveIDs := lo.Map(ghsas, func(ghsa githubSecurityAdvisoryResponse, _ int) string {
```
pdeslaur, line 63: Should this variable be named `ghsa`?

pkg/advisory/alias_finder.go
luhring: Tell me more about that idea :) We also have the option to return an `error` here, by the way, it's just this error case didn't seem critical to the logic flow here. Are you thinking panicking would be especially useful here?

pkg/advisory/alias_finder.go
luhring: Good callout, I actually started with trying to use https://github.com/google/go-github, but then I saw they don't support the endpoint I need just yet. There's an open issue for it: https://github.com/google/go-github/issues/2851

pkg/advisory/alias_finder.go
luhring: Good catch! Fixed!

pkg/advisory/discover_aliases.go
```diff
@@ -0,0 +1,150 @@
...
+// documents with the discovered aliases.
+func DiscoverAliases(ctx context.Context, opts DiscoverAliasesOptions) error {
+	documents := opts.AdvisoryDocs.Select().Configurations()
+
+	// If the caller selected specific packages (by name), filter the set of
+	// advisory documents to just those packages.
+	if len(opts.SelectedPackages) > 0 {
+		documents = lo.Filter(documents, func(doc v2.Document, _ int) bool {
+			return slices.Contains(opts.SelectedPackages, doc.Package.Name)
+		})
```
pdeslaur, line 41: We'd have an opportunity to make `opts.SelectedPackages` a set instead of a slice. Probably not a problem right now, but if we have many packages, that could be slow.

pkg/advisory/discover_aliases.go
```diff
@@ -0,0 +1,150 @@
...
+
+	for _, doc := range documents {
+		for _, adv := range doc.Advisories {
+			// If the advisory ID is already a CVE ID, do a lookup and ensure the discovered
+			// aliases are present in the advisory's list of aliases.
+			if vuln.RegexCVE.MatchString(adv.ID) {
+				ghsas, err := opts.AliasFinder.GHSAsForCVE(ctx, adv.ID)
+				if err != nil {
+					return err
+				}
```
pdeslaur, line 52: If I'm reading this correctly, we'll send a GitHub request for each CVE. This may be premature optimization, but in some cases we have a CVE showing up in many packages. In that case, we'd end up requesting the same information from GitHub repeatedly.

pkg/advisory/discover_aliases.go
```diff
@@ -0,0 +1,150 @@
...
+		documents = lo.Filter(documents, func(doc v2.Document, _ int) bool {
+			return slices.Contains(opts.SelectedPackages, doc.Package.Name)
+		})
+	}
+
+	for _, doc := range documents {
+		for _, adv := range doc.Advisories {
+			// If the advisory ID is already a CVE ID, do a lookup and ensure the discovered
+			// aliases are present in the advisory's list of aliases.
+			if vuln.RegexCVE.MatchString(adv.ID) {
```
pdeslaur, line 48: I wonder if we could extract the logic for each of the `if` statements in functions. It gets a bit hard to read!

This pattern also reminds me of the switch with no condition, e.g. this might read better:

```
switch {
case vuln.RegexCVE.MatchString(adv.ID):
    // logic
case vuln.RegexGHSA.MatchString(adv.ID):
    // logic
default:
    // whatever
}
```

pdeslaur (commented):
Mostly done reviewing. Many of the comments are superficial / growth concerns - we can find a reasonable balance and get this out the door tomorrow

pkg/advisory/discover_aliases.go
luhring: That's right. I had mentioned this in the PR description as a potential enhancement:

>6. Cache API queries in memory during command execution, since there's a decent amount of redundancy today of vulnerability IDs in our advisory data set.

Would following up with this optimization be okay with you? Or would you rather see it get in immediately?

pkg/advisory/discover_aliases.go
luhring: Great suggestion! Updated!

pkg/advisory/discover_aliases.go
luhring: Done!

pkg/advisory/alias_finder.go
pdeslaur: Oh, I didn't read that code carefully enough - we're trying to read the body in case of an error - disregard!

pkg/advisory/discover_aliases.go
pdeslaur: Ah - all good. Let's keep that as a future improvement

pdeslaur (approved):

PR #429, opened 2023-10-16
More documentation for advisory commands

luhring:
This PR adds more documentation and help text for the `advisory` commands, including a new draft of a document explaining how the "distro detection" mechanism works today.

This also makes other polishing tweaks:

- Renames the `wolfictl adv db` command to `wolfictl adv secdb` — because soon we'll want to produce other kinds of databases besides the Alpine-style "secdb". For now, `db` is kept as an alias to `secdb` to avoid any immediate breakages.
- Adds the `ls` alias to `wolfictl adv list`
- Marks the `migrate` command as deprecated — we'll remove it in another PR soon


cpanato (approved):
🎉

PR #433, opened 2023-10-17
feat(scan): ability to base scan(s) on a Melange build log

luhring:
## Background

In some workflows, when we build packages, we ask Melange to create a "build log" of the packages that it outputs. (This can be accomplished with the `melange build` flag called `--create-build-log`.)

## What's new

This PR adds an enhancement to the `wolfictl scan` command to determine the set of APKs to scan off the build log, so we can scan _N_ packages that we just built.

For example, I just did a Melange build of the "jansson" package, which also has 1 subpackage:

```console
$ cat packages.log 
aarch64|jansson|jansson|2.14-r0
aarch64|jansson|jansson-dev|2.14-r0
$ wolfictl scan --build-log .                                        
🔎 Scanning "packages/aarch64/jansson-2.14-r0.apk"
✅ No vulnerabilities found
🔎 Scanning "packages/aarch64/jansson-dev-2.14-r0.apk"
✅ No vulnerabilities found
```

The new `--build-log` flag requires that the command is only given a single argument. This argument can be either a filepath to a build log, or a path to a directory that contains the build log as `packages.log`.

In a future PR, it'd be great to catch up `wolfictl sbom` to support all the kinds of input that `wolfictl scan` supports now.

Big shout out to @rawlingsj for thinking of this feature! 🎉

### DIFF
--- a/pkg/buildlog/buildlog.go
+++ b/pkg/buildlog/buildlog.go
@@ -1,0 +1,45 @@
+package buildlog
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"strings"
+)
+
+// Entry represents a single line in a Melange build log.
+type Entry struct {
+	Arch, Origin, Package, FullVersion string
+}
+
+// Parse parses a Melange build log into a slice of entries.
+func Parse(r io.Reader) ([]Entry, error) {
+	splitFunc := func(ru rune) bool {
+		return string(ru) == "|"
+	}
+
+	var entries []Entry
+
+	scanner := bufio.NewScanner(r)
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		fields := strings.FieldsFunc(line, splitFunc)
+		if len(fields) != 4 {
+			return nil, fmt.Errorf("invalid line %q, expected 4 '|'-delimited fields", line)
+		}
+
+		entry := Entry{
+			Arch:        fields[0],
+			Origin:      fields[1],
+			Package:     fields[2],
+			FullVersion: fields[3],
+		}
+		entries = append(entries, entry)
+	}
+	if err := scanner.Err(); err != nil {
+		return nil, fmt.Errorf("scanning lines of build log: %w", err)
+	}
+
+	return entries, nil
+}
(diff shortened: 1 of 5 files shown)


pkg/cli/scan.go
```diff
@@ -127,8 +160,103 @@ func cmdScan() *cobra.Command {
...
+
+	return is, nil
+}
+
+// resolveInputFilePathsFromBuildLog takes the given path to a Melange build log
+// file (or a directory that contains the build log as a "packages.log" file).
+// Once it finds the build log, it parses it, and returns a slice of file paths
+// to APKs to be scanned. Each APK path is created with the assumption that the
+// APKs are located at "./packages/$ARCH/$PACKAGE-$VERSION.apk".
+func resolveInputFilePathsFromBuildLog(buildLogPath string) ([]string, error) {
```
luhring, line 209: 🤔 Would it be better to create these APK file paths relative to `filepath.Dir(buildLogPath)`, rather than always relative to `.`?

pkg/buildlog/buildlog.go
```diff
@@ -0,0 +1,45 @@
...
+		return string(ru) == "|"
+	}
+
+	var entries []Entry
+
+	scanner := bufio.NewScanner(r)
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		fields := strings.FieldsFunc(line, splitFunc)
```
imjasonh, line 27: Why this instead of `strings.Fields(line, "|")` ?

imjasonh (approved):

pkg/buildlog/buildlog.go
luhring: Maybe I missed it — is that a thing? https://pkg.go.dev/strings#Fields

pkg/cli/scan.go
luhring: We can follow up if this turns out to be a counterintuitive behavior as-is

pkg/buildlog/buildlog.go
luhring: I'll resolve for now but I can follow up with a PR if there's a better way here!

PR #437, opened 2023-10-17
add option to set a custom path

cpanato:
add an option to set a custom temp path to store the apk for the sbom generation

this is required for our cve automation because we have a rule that denies the usage of the `/tmp` directory in a k8s deployments.

### DIFF
--- a/pkg/cli/scan.go
+++ b/pkg/cli/scan.go
@@ -216,14 +216,14 @@ func scanInput(inputFilePath string, p *scanParams) (*inputScan, error) {
 }
 
 type scanParams struct {
+	disableSBOMCache    bool
+	sbomInput           bool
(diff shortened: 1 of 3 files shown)


luhring:
@cpanato Can we control the tempdir location using the `TMPDIR` environment variable here? That's usually how I see this kind of thing handled, unless it doesn't apply for some reason

vaikas (approved):

cpanato:
> eant to say is that Go and the operating system handle this automatically for us — depending on the exact circumstance. Without us having to do anything special, if the Go program sees a path set in the environment variable `TMPDIR`, the standard library will use that for function calls like `os.MkdirTemp`.

oh i see, i did not know that, TIL
thanks dan

PR #438, opened 2023-10-18
wolfictl scan: various improvements

luhring:
This PR makes the following improvements to `wolfictl scan`:

- When using `--build-log`, APK package paths are now resolved relative to the provided build log path, rather than to the current working directory (since build logs and `packages/<arch>/*.apk` files are virtually always colocated)
- When using `--require-zero` to exit 1 on a failed scan, we now allow all scans to complete before exiting, rather than exiting on the first failed scan in a group
- You can now use the `-a` flag multiple times to leverage advisory data from multiple local directories at the same time (when filtering scan results)

### DIFF
--- a/pkg/scan/filter.go
+++ b/pkg/scan/filter.go
@@ -16,74 +16,93 @@ const (
 var ValidAdvisoriesSets = []string{AdvisoriesSetResolved, AdvisoriesSetAll}
 
 // FilterWithAdvisories filters the findings in the result based on the advisories for the target APK.
-func FilterWithAdvisories(result *Result, advisoryCfgs *configs.Index[v2.Document], advisoryFilterSet string) ([]*Finding, error) {
+func FilterWithAdvisories(result *Result, advisoryDocIndices []*configs.Index[v2.Document], advisoryFilterSet string) ([]*Finding, error) {
 	if result == nil {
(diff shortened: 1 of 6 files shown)


pkg/scan/filter.go
```diff
@@ -16,74 +16,93 @@ const (
...
+	// TODO: Should we error out if we end up with multiple documents for a single package?
+
 	if len(documents) == 0 {
 		// No advisory configs for this package, so we know we wouldn't be able to filter anything.
 		return result.Findings, nil
 	}
 
-	// We know there's an advisories document for this package, so we can get the advisories.
-	packageAdvisories := documents[0].Advisories
+	// Use a copy of the findings so we don't mutate the original result.
```
jonjohnsonjr, line 41: https://pkg.go.dev/golang.org/x/exp/slices#Clone

imjasonh (approved):

pkg/scan/filter.go
luhring: Does that do the exact same thing?

jonjohnsonjr: I think yes! Either one sgtm.

cpanato (approved):
looks cool

jonjohnsonjr (approved):

PR #440, opened 2023-10-22
grype: update dependency and match matchers

ajayk:


luhring:
For reference: https://github.com/anchore/grype/pull/1565

luhring (approved):
Thanks @ajayk!

PR #443, opened 2023-10-24
look for GH tokens used by gh CLI

imjasonh:
This change makes it so you don't need a `GITHUB_TOKEN` env var set if you've previously set up and authorized [GitHub's `gh` CLI](https://cli.github.com/) -- if you have, the token used by that CLI will be used.

[`TokenForHost`](https://pkg.go.dev/github.com/cli/go-gh/v2/pkg/auth#TokenForHost) checks `GITHUB_TOKEN` and falls back to a plaintext file or secure keyring storage.

### DIFF
--- a/pkg/cli/gh_gc_branch.go
+++ b/pkg/cli/gh_gc_branch.go
@@ -2,20 +2,20 @@ package cli
 
 import (
 	"context"
+	"errors"
 	"log"
-	"os"
 	"strings"
 	"time"
 
+	ghauth "github.com/cli/go-gh/v2/pkg/auth"
 	"github.com/google/go-github/v55/github"
-
-	http2 "github.com/wolfi-dev/wolfictl/pkg/http"
+	"github.com/spf13/cobra"
 	"golang.org/x/oauth2"
 	"golang.org/x/time/rate"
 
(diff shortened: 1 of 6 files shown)


pkg/cli/gh_gc_branch.go
```diff
@@ -53,6 +50,15 @@ wolfictl gh gc branch https://github.com/wolfi-dev/os --match "wolfictl-"
 	return cmd
 }
 
+type ghTokenSource struct{}
+
+func (ghTokenSource) Token() (*oauth2.Token, error) {
+	if tok, _ := ghauth.TokenForHost("github.com"); tok != "" {
+		return &oauth2.Token{AccessToken: tok}, nil
+	}
```
rawlingsj, line 58: should we have a fall back of checking if the `GITHUB_TOKEN` env var is set so that we don't break existing approaches that use the env var approach?

rawlingsj (approved):

pkg/cli/gh_gc_branch.go
imjasonh: `TokenForHost` checks the env var first and falls back.

cpanato (approved):

PR #446, opened 2023-10-26
Advisory schema v2.0.1: new PendingUpstreamFix event type

luhring:
This revs the advisory schema to version `2.0.1`, which adds a non-breaking change:

- A new event type `PendingUpstreamFix`

This addresses an emergent need where the distro staff either isn't able to implement the fix on the upstream maintainers' behalf in a reasonable amount of time, or where doing so introduces more risk to the security of the software than leaving the fix to the upstream maintainers to take on.

This is a distinct state from `FixNotPlanned`, which says that a fix is never coming.

It's also distinct from `TruePositiveDetermination`, which acknowledges the vulnerability's impact on the package but doesn't communicate the distro staff's plan for remediation.

### DIFF
--- a/pkg/advisory/create.go
+++ b/pkg/advisory/create.go
@@ -57,6 +57,12 @@ func Create(req Request, opts CreateOptions) error {
 			return fmt.Errorf("unable to create advisory %q for %q: %w", newAdvisoryID, req.Package, err)
 		}
 
+		// Update the schema version to the latest version.
+		err = documents.Update(v2.NewSchemaVersionSectionUpdater(v2.SchemaVersion))
+		if err != nil {
+			return fmt.Errorf("unable to update schema version for %q: %w", req.Package, err)
+		}
+
 		return nil
 	}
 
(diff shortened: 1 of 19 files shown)


hectorj2f (dismissed):

luhring:
Perhaps the `PendingUpstreamFix` type should join the group of event types that are filtered out when doing a `wolfictl scan` with the `resolved` filter 🤔

pkg/advisory/create.go
```diff
@@ -57,6 +57,12 @@ func Create(req Request, opts CreateOptions) error {
 			return fmt.Errorf("unable to create advisory %q for %q: %w", newAdvisoryID, req.Package, err)
 		}
 
+		// Update the schema version to the latest version.
+		err = documents.Update(v2.NewSchemaVersionSectionUpdater(v2.SchemaVersion))
+		if err != nil {
+			return fmt.Errorf("unable to update schema version for %q: %w", req.Package, err)
+		}
```
luhring, line 64: This stanza is repeated a few times throughout the PR diff. It should be consolidated into one location — and that's not easy to do right now. We're due for a small refactor in how we interact with the underlying data, at which point this will get better.

pdeslaur (approved):

PR #450, opened 2023-10-29
feat(scan): experimental `--govulncheck` flag for triaging

luhring:
Adds a new **experimental** flag `--govulncheck` to the `wolfictl scan` command.

This feature integrates with [govulncheck](https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck) (as a library, so no extra installation is needed), to follow up the Grype scan with an extra close inspection of the APK's _Go binaries_ that were referenced by the vulnerability matching output from Grype.

Govulncheck is great because it crosschecks binaries' symbol table data with symbol-level vulnerability data in the Go Vuln DB, which helps users understand if a specific symbol (function/method) that's known to be vulnerable is actually present in a compiled Go binary. This level of granularity can help confirm that an initially reported vulnerability match (typically at the coarser "Go module" level) is a true positive (the affected function is included) or a false positive (the affected function isn't found in the compiled binary).

With respect to `wolfictl scan`, this PR introduces the beginnings of a _"triaging framework"_, where reported findings can be assessed by one or more triage sources (like "govulncheck") as a true or false positive. This framework is expected to evolve heavily over time and should be considered unstable for the time being.

image

When `--govulncheck` is used, triage assessments are added to the default (tree) command output, and a bit more information is found in the JSON (`-o json`) output as well.

**NOTE:** One useful aspect of this feature **that isn't included in this PR** (but would be great to add soon) is to assess a finding as a false positive when the match is in a dependency module and the vulnerability has been **excluded** from the Go Vuln DB on the grounds that the affected functionality is "NOT_IMPORTABLE". These entries are not included in the Go Vuln DB, so adding this feature to `wolfictl` will require creating an additional data source.

Feedback on this feature is very much welcomed!

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -46,6 +46,7 @@ require (
 	golang.org/x/sync v0.4.0
 	golang.org/x/text v0.13.0
 	golang.org/x/time v0.3.0
+	golang.org/x/vuln v1.0.1
 	gopkg.in/yaml.v3 v3.0.1
 	sigs.k8s.io/release-utils v0.7.5
 )
@@ -318,3 +319,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.3.0 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace golang.org/x/vuln => github.com/luhring/golang-vuln v1.0.2-0.20231029212121-c364fd4725dc
--- a/pkg/cli/scan.go
+++ b/pkg/cli/scan.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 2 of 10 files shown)


go.mod
```diff
@@ -318,3 +319,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.3.0 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace golang.org/x/vuln => github.com/luhring/golang-vuln v1.0.2-0.20231029212121-c364fd4725dc
```
hectorj2f, line 323: Is this simply a fork or we need some changes in there ?

luhring: Yeah, it's a fork that has changes we need. My hope is to figure out if/how we can get these upstreamed!

hectorj2f: I'm wondering if it makes sense to move it under the `chainguard-dev` repositories.

pkg/cli/scan.go
```diff
@@ -343,6 +374,7 @@ func (p *scanParams) addFlagsTo(cmd *cobra.Command) {
 	cmd.Flags().StringVarP(&p.advisoryFilterSet, "advisory-filter", "f", "", fmt.Sprintf("exclude vulnerability matches that are referenced from the specified set of advisories (%s)", strings.Join(scan.ValidAdvisoriesSets, "|")))
 	cmd.Flags().StringSliceVarP(&p.advisoriesRepoDirs, "advisories-repo-dir", "a", nil, "local directory for advisory data")
 	cmd.Flags().BoolVar(&p.disableSBOMCache, "disable-sbom-cache", false, "don't use the SBOM cache")
+	cmd.Flags().BoolVar(&p.triageWithGoVulnCheck, "govulncheck", false, "EXPERIMENTAL: triage vulnerabilities in Go binaries using govulncheck")
```
hectorj2f, line 377: suggestion: Perhaps we prefer to mark this flag as hidden, as long as it is experimental.

hectorj2f (dismissed):
LGTM, I just left two minor comments.

pkg/scan/apk.go
```diff
@@ -141,15 +135,18 @@ func scan(s *sbomSyft.SBOM, localDBFilePath string) (*Result, error) {
...
+	var findings []Finding
 	for i := range matches {
 		m := matches[i]
 
 		finding, err := mapMatchToFinding(m, datastore)
 		if err != nil {
 			return nil, fmt.Errorf("failed to map match to finding: %w", err)
 		}
-		findings = append(findings, finding)
+		if finding == nil {
```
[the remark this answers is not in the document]
luhring: Not today, but wanted to be safe here to avoid a panic in the future from changes to the mapping logic, since we're dereferencing the pointer immediately after here

pkg/cli/scan.go
luhring: Sure, updated 👍

go.mod
luhring: Sure, we can follow up on that if it's important. It wasn't obvious to me that that's the right call here for a community-owned repo

hectorj2f (approved):

PR #454, opened 2023-10-30
Fix linter parity with local development

luhring:
We were seeing different results in CI vs. the local experience — including a case where applying a local "fix" _breaks_ the results in CI — which makes it painful to run the check locally. Ideally running the local check rewards developers by shifting "left" feedback that would come back in CI to the local iteration workflow.

This PR updates CI to use the latest version of `golangci-lint` and fixes the new linting issues that get reported.

### DIFF
--- a/.github/workflows/golangci-lint.yaml
+++ b/.github/workflows/golangci-lint.yaml
@@ -18,4 +18,4 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@3a919529898de77ec3da873e3063ca4b10e7f5cc # v3.7.0
         with:
-          version: v1.54
+          version: v1.55.1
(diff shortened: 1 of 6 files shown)


.github/workflows/golangci-lint.yaml
```diff
@@ -18,4 +18,4 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@3a919529898de77ec3da873e3063ca4b10e7f5cc # v3.7.0
         with:
-          version: v1.54
+          version: v1.55.1
```
cpanato, line 21:
```suggestion
          version: v1.55
```

then it will pick that patch releases automatically and we dont need to bump everytime

luhring: I'm cool with that too! As long as the local results match CI's results

PR #459, opened 2023-11-03
perf: cache results for GHSA queries

luhring:
In the [PR that originally introduced vulnerability alias discovery](https://github.com/wolfi-dev/wolfictl/pull/419), a noted "future work" item was to cache results from lookups of CVE and GHSA IDs using the GitHub API.

This PR introduces a simple in-memory caching mechanism to avoid redundant calls to the GitHub API.

**Sample runs of alias discovery for a distro:**

Before:

```
wolfictl adv alias discover  2.10s user 0.91s system 5% cpu 56.769 total
```

After:

```
wolfictl adv alias discover  1.45s user 0.36s system 11% cpu 15.502 total
```

### DIFF
--- a/pkg/advisory/alias_finder.go
+++ b/pkg/advisory/alias_finder.go
@@ -19,16 +19,25 @@ type AliasFinder interface {
 }
 
 type HTTPAliasFinder struct {
-	client *http.Client
+	client          *http.Client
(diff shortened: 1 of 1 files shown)


cpanato (approved):
just need to fix the lint

cpanato (approved):

Issue #468, opened 2023-11-07
Use `REGEX` instead of `strings.Containes` in `tag-filter`


debasishbsws:
In a current change #458, `tag-filters` doesn't match only with the START of the string value. And it is useful for something like `openjdk*-ga` releases.

But the problem is with some of the current wolfi packages that point to a specific version.
For example:
bazel-3, bazel-5 etc https://github.com/wolfi-dev/os/pull/8094
cython-0 https://github.com/wolfi-dev/os/pull/8093

Though we have ignore-regex-patterns now as mentioned in the [previous PR](https://github.com/wolfi-dev/wolfictl/pull/458#:~:text=Important%3A%20this%20could%20affect%20existing%20tag%2Dfilter%20usage%20where%20we%20would%20need%20to%20add%20ingore%2Dregex%20config%20to%20packages%20that%20use%20a%20filter%20and%20rely%20on%20matching%20only%20the%20START%20of%20the%20string%20value.%20I%27m%20running%20tests%20against%20my%20fork%20of%20wolfi%20to%20identify%20these%20cases%20and%20will%20fix%20any.). Adding negation regex is relatively more complex and ERROR prone.
e.g: for `bazel-3` we need to add 
```yaml
ignore-regex-patterns:
    - '^(?!3\.).*'
```

### Solution:
instead of strings.Containes we can use regex.
```golang
	// the github graphql query filter matches any occurrence of the tag filter
	if ghm.TagFilter != "" {
		regex, err := regexp.Compile(ghm.TagFilter)
		if err != nil {
			return "", errors.Wrapf(err, "failed to compile regex %s", ghm.TagFilter)
		}
		if !regex.MatchString(v) {
			return "", nil
		}
	}
```

- this will not change the current functionality, because without special chars like '^' or '$', it will work as same as `strings.Containes` function.
- It will reduce the complexity of writing negative regex(more complex). And we can just change the tag filter from "3." to "^3." in [bazel-3](https://github.com/wolfi-dev/os/blob/5754ec35e16397ad7746cda1202b525f6c2b9972/bazel-3.yaml#L74), cython-0 and any other necessary.

/cc @rawlingsj

rawlingsj:
I like this a lot. Indeed it was an oversight on my behalf so thanks for looking at this as well as the time you've clearly taken to understand the reasons behind the initial change!

/cc @ajayk as he also saw this issue.

debasishbsws:
[...]

PR #469, opened 2023-11-07
Use `REGEX` instead of `strings.Containes` to match  `tag-filter`

debasishbsws:
Fixes: #468 

/cc @rawlingsj @ajayk


ajayk:
Thanks @debasishbsws

ajayk (approved):

PR #477, opened 2023-11-09
feature: add tag-filter-prefix and tag-filter-contains

ajayk:
This PR introduces two new filters  **`tag-filter-prefix`** and **`tag-filter-contains`**, to the GitHub Release filtering . These features enhance the filtering capabilities when preparing versions.

 `tag-filter-prefix` option ensures that only tags that start with the specified prefix are considered. the functionality is same as the `tag-filter`

`tag-filter-contains` option ensures that only tags containing the specified substring are considered.

**``deprecation-notice``** : we would like to deprecate `tag-filter` in favor of `tag-filter-prefix`  , once all the configurations are moved over , we will remove supporting `tag-filter`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,11 +4,11 @@ go 1.21.1
 
 require (
 	chainguard.dev/apko v0.11.2
-	chainguard.dev/melange v0.5.2-0.20231027223504-2bcfbe6439a5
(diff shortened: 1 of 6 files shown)


cpanato (approved):
Lgtm

PR #500, opened 2023-11-20
lint: detect url mimicry attacks

tstromberg:
If someone adds or updates a package with a very similar hostname to other packages, complain because such an inconsistency is unlikely to be wanted or legitimate.

This PR adds detection if a package is updated or added that falls into one of two categories:

- "openssl.com" instead of "openssl.org"
- "libshh2.org" instead of "libssh2.org"

Unfortunately, this will not detect if *all* references to a domain are updated to a domain under the control of an attacker, as the lint tool only has access to the current repository state.

This PR does introduce two tangentially unrelated changes due to issues I ran into during testing:

- Stable sort order for test comparisons
- A null reference panic if you pass a nonexistent directory to `ReadAllPackagesFromRepo()`

One interesting thing is that this PR opens the door up to checks that validate packages against other packages. This first iteration is doing so through a global, but once we've proven the approach, we should pass structs around to make it more scalable.

Related PR: #496


rawlingsj (approved):

rawlingsj:
This is great!

PR #507, opened 2023-12-02
chore: have dependabot ignore syft+grype

luhring:
(Reasons for this are documented as comments in the dependabot configuration)

This is intended to be temporary — trading off a small amount of manual work for improved stability and reliability of our scanning.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,3 +16,6 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    ignore:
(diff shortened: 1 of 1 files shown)


cpanato (commented):
will we check those manually and upgrade as we wish?

luhring:
> will we check those manually and upgrade as we wish?

Yeah exactly — but whoever's upgrading them should make sure to read the release notes for the new versions and ensure there that any breaking changes are accounted for on our side.

hectorj2f (approved):
lgtm

PR #513, opened 2023-12-08
feat(adv): v2.0.2 schema

luhring:
This PR adds a new non-breaking ("addition") advisory schema update: **v2.0.2**.

This adds support for capturing valuable information from a vulnerability scan into an advisory's **detection** event(s).

This also tests out the idea of "versioned" types in the advisory schema, here by introducing a new detection type of `scan/v1`. This is intended to allow us to iterate on the data shape such that multiple versions of this detection type can coexist in a single advisory data set, and newer detections can always take advantage of the latest data structure available.

The hope is that by adding this library support, we can begin automatically generating advisory data using vulnerability scans, such that we're not losing the insightful data found at scan-time as we encode the detection into our advisory data store.

**Note:** This PR doesn't add any actual exercising of the new schema yet. At some point relatively soon, it'd be neat to add an option to wolfictl's scanning to immediately record net-new vulnerabilities using this new detection type.

cc: @rawlingsj @cpanato @pdeslaur

### DIFF
--- a/pkg/configs/advisory/v2/detection.go
+++ b/pkg/configs/advisory/v2/detection.go
@@ -14,6 +14,7 @@ import (
 const (
 	DetectionTypeManual = "manual"
 	DetectionTypeNVDAPI = "nvdapi"
+	DetectionTypeScanV1 = "scan/v1"
 )
 
 var (
@@ -21,6 +22,7 @@ var (
 	DetectionTypes = []string{
 		DetectionTypeManual,
 		DetectionTypeNVDAPI,
+		DetectionTypeScanV1,
 	}
 )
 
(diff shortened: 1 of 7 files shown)


pkg/configs/advisory/v2/detection.go
```diff
@@ -124,3 +136,51 @@ func (d DetectionNVDAPI) Validate() error {
...
+	ComponentType     string `yaml:"componentType"`
+	ComponentLocation string `yaml:"componentLocation"`
+	Scanner           string `yaml:"scanner"` // TODO: it'd be nice for the scanner value to be automatically versioned
+}
+
+// Validate returns an error if the DetectionScanV1 data is invalid.
+func (d DetectionScanV1) Validate() error {
+	// TODO: Should SubpackageName be required, and it's set to the origin package
+	// 	sometimes? Or should an empty value imply this was a scan of an origin
+	// 	package?
```
luhring, line 163: Surfacing this question for PR discussion!

pdeslaur: I'd require it for consistency but I don't have a really strong opinion!

rawlingsj: yeah let's always set it

luhring: Done!

rawlingsj (approved):

Issue #520, opened 2023-12-13
wolfictl scan should initialize the grype DB once


jonjohnsonjr:
**Description**

This line is particularly expensive: https://github.com/wolfi-dev/wolfictl/blob/33180c3209838e05572c4fd044dbe53ca151ba07/pkg/scan/apk.go#L120

It's actually twice as expensive as it _could_ be: https://github.com/anchore/grype/issues/1502

But the main problem is that we call it for every package we're scanning instead of doing it once per `wolfictl` invocation.

As a quick win, I'm going to drop the `ValidateByHashOnGet` from `grypeDBConfig`, but that's a little janky.

An example of this is that scanning all the `glibc` and `busybox` packages took ~15 minutes in CI: https://github.com/wolfi-dev/os/actions/runs/7199389272/job/19611327817?pr=9871

PR #521, opened 2023-12-13
Don't validate grype DB by hash

jonjohnsonjr:
This is very expensive and we do it for every APK that gets scanned.

We should consider re-enabling this if we initialize the DB once per invocation instead of per APK.

X-ref: https://github.com/wolfi-dev/wolfictl/issues/520


jonjohnsonjr:
Overall time for all the `glibc` packages drops from ~5 minutes to ~1.5 minutes.

You can see in Before that we spend most of the time in sha256 verifying the DB.

Before: 
image

After: 
image

PR #522, opened 2023-12-13
Cache grype DB initialization across scans

jonjohnsonjr:
We were spending most of our time hashing the grype DB instead of scanning things. This introduces a scan.Scanner to hold the cached DB so we spend (a lot) more time scanning.

Fixes https://github.com/wolfi-dev/wolfictl/issues/520


jonjohnsonjr:
Compare to alternative: https://github.com/wolfi-dev/wolfictl/pull/521#issuecomment-1854549853

image

luhring (approved):

PR #524, opened 2023-12-14
Scan packages in parallel

jonjohnsonjr:
Instead of scanning one package at a time, scan GOMAXPROCS at a time.

### DIFF
--- a/pkg/cli/scan.go
+++ b/pkg/cli/scan.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"strings"
 
@@ -27,6 +28,7 @@ import (
 	"github.com/wolfi-dev/wolfictl/pkg/scan"
 	"github.com/wolfi-dev/wolfictl/pkg/versions"
 	"golang.org/x/exp/slices"
+	"golang.org/x/sync/errgroup"
 )
 
 const (
@@ -35,11 +37,6 @@ const (
 )
 
 var validOutputFormats = []string{outputFormatOutline, outputFormatJSON}
-
-type inputScanResult struct {
-	InputFile string
-	Result    scan.Result
-}
 
 func cmdScan() *cobra.Command {
 	p := &scanParams{}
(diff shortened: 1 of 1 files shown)


pkg/cli/scan.go
```diff
@@ -205,19 +208,36 @@ wolfictl scan package1 package2 --remote
...
-					// Accumulate the list of failures to be returned at the end, but we still want to complete all scans
-					inputPathsFailingRequireZero = append(inputPathsFailingRequireZero, input)
-				}
+				g.Go(func() error {
+					isr, err := p.doScanCommandForSingleInput(cmd.Context(), input, advisoryDocumentIndices)
+					if err != nil {
+						return err
+					}
+
+					scans[i] = *isr
```
luhring, line 225: Are we sure this kind of slice access is thread-safe?

pkg/cli/scan.go
```diff
@@ -205,19 +208,36 @@ wolfictl scan package1 package2 --remote
...
-				scans = append(scans, *isr)
+			for i, input := range inputs {
+				i, input := i, input
 
-				if p.requireZeroFindings && len(isr.Result.Findings) > 0 {
-					// Accumulate the list of failures to be returned at the end, but we still want to complete all scans
-					inputPathsFailingRequireZero = append(inputPathsFailingRequireZero, input)
-				}
+				g.Go(func() error {
+					isr, err := p.doScanCommandForSingleInput(cmd.Context(), input, advisoryDocumentIndices)
```
luhring, line 220: Actually, have we checked that everything called within this point is thread-safe? There's a ton happening here. Just curious!

pkg/cli/scan.go
jonjohnsonjr: Yep! Appending isn't, which is why below on findings I use the lock to append to `inputPathsFailingRequireZero`.

pkg/cli/scan.go
jonjohnsonjr: Sure. I've kept the actually scanning bits serial, but started fanning out the SBOM generation and also overlapped that with the vulnerability database loading.

jonjohnsonjr:
Before: 
image

After: 
image

pdeslaur (approved):

PR #537, opened 2024-01-02
feat: wolfictl gc issues

rawlingsj:
example:
```sh
wolfictl gc issues https://github.com/wolfi-dev/versions --match "version-stream:"
```


pkg/cli/gh_gc_issues.go
```diff
@@ -0,0 +1,87 @@
...
+			return gcIssues(client, args[0], match)
+		},
+	}
+
+	cmd.Flags().StringVar(&match, "match", "", "pattern to match issues against")
+
+	return cmd
+}
+
+func gcIssues(ghclient *http2.RLHTTPClient, repo, match string) error {
```
luhring, line 54: question: does it make any sense to verify that `match` is non-empty before using it in `HasPrefix`, just to avoid mass closing of issues?

rawlingsj: SGTM

imjasonh (approved):

PR #557, opened 2024-01-17
bump deps, use slog and clog

imjasonh:
This picks up recent changes to apko, melange and go-apk to use `log/slog` structured logging, and `github.com/chainguard-dev/clog` to pass around a logger with the context.

By standardizing on this form, if we move any part of this code to run in a cloud service, we can have those structured logs also get slurped into GCP logs in the standard way using https://github.com/imjasonh/gcpslog


luhring (approved):
you're awesome

PR #559, opened 2024-01-18
chore: adapt to breaking changes in syft

luhring:
This handles the update to:
- syft @ v0.101.0
- grype @ v0.74.1


luhring:
I'm mostly sure I did this correctly 😅

pdeslaur (approved):

hectorj2f:
In the future, we should add tests to validate any changes done here.

PR #561, opened 2024-01-19
release: don't bump tags if no changes

imjasonh:
Some releases are redundant, since no changes happen during the intervening time. We should try to avoid these.

This is an attempt to avoid these.

Screenshot 2024-01-19 at 11 28 45 AM


luhring (approved):

imjasonh:
Example from my fork: https://github.com/imjasonh/wolfictl/actions/runs/7586699145/job/20665440184

PR #562, opened 2024-01-19
feat: new command "image apk"

luhring:
This adds a new experimental group of subcommands to wolfictl, `wolfictl image ...`, and a subcommand that's useful for vulnerability triaging called `wolfictl image apk`.

I'm not 100% sure we want image-related functionality in `wolfictl`, but after talking it over with @jonjohnsonjr for a while, it seems like not the worst place to house an experiment like this, for Wolfi/apk-centric operations that are specific to the container image use case of the distro. If others have thoughts/feelings about this, I'd like to hear it!

The new `wolfictl image apk` command shows what APKs are included in an image, with variations that start to make manual triaging and remediation a bit less painful.

For example, before, to find out which part of an image had a component (e.g. Go module) in question, we'd end up processing Grype and Syft JSON, comparing artifact IDs, and searching through an array of relationships. This new operation lets you do all that in one command.

This also lets you supply one or more directories that contain Melange configs, so you can quickly navigate to the file where you'd want to start iterating on fixes/adjustments to APK packages.

Here's the full help for the command.

```console
$ wolfictl image apk -h
Show APK(s) in a container image

Usage:
  wolfictl image apk <image> [flags]

Examples:

  # Show all APKs in an image
  wolfictl image apk cgr.dev/chainguard/bash

  # Show all APKs in an image that own a component (based on a Syft analysis)
  wolfictl image apk cgr.dev/chainguard/coredns -c 'github.com/aws/aws-sdk-go'

  # Show all APKs in an image that own a component, and show the path to the
  # Melange configuration file for each APK, within the given directory
  wolfictl image apk cgr.dev/chainguard/prometheus-operator -c 'github.com/aws/aws-sdk-go' -d ~/code/wolfi-os


Flags:
  -c, --component string     show only APKs containing the given component
  -d, --distro-dir strings   path to a directory containing Melange build configuration files
  -h, --help                 help for apk
```


pkg/cli/image_apk.go
```diff
@@ -0,0 +1,245 @@
...
+			return nil, fmt.Errorf("unable to index configuration files from %q: %w", p, err)
+		}
+		configIndexes = append(configIndexes, index)
+		indexRootPaths = append(indexRootPaths, p)
+	}
+
+	configPaths := make([]string, 0, len(apks))
+
+	for i := range apks {
+		apk := apks[i]
```
pdeslaur, line 218: nit: 
```suggestion
	for i, apk := range apks {
```

pkg/cli/image_apk.go
```diff
@@ -0,0 +1,245 @@
...
+		apkPkgID := rel.From.ID()
+		ownedPkgID := rel.To.ID()
+
+		apkPkg := r.sbom.Artifacts.Packages.Package(apkPkgID)
+		if apkPkg == nil {
+			return fmt.Errorf("unable to find owner package %q", apkPkgID)
+		}
+		if apkPkg.Type != pkg.ApkPkg {
+			return fmt.Errorf(
+				"expected APK package as owner in relatioship, got %s (for package %q)",
```
pdeslaur, line 179:
```suggestion
				"expected APK package as owner in relationship, got %s (for package %q)",
```

pkg/cli/image_apk.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+func (r *syftResults) apksOwningPackageWithName(name string) ([]pkg.Package, error) {
+	// Lazily create the index of owned package (by name) to APK package
+	if r.packageOwnershipsByOwnedPackageName == nil {
+		err := r.indexPackageOwnerships()
+		if err != nil {
+			return nil, fmt.Errorf("unable to index APK package ownerships: %w", err)
+		}
+	}
```
pdeslaur, line 145: nit: Not very important - but I think it'd be cleaner to always call `indexPackageOwnerships()`, and have that method return `r.packageOwnershipsByOwnedPackageName`. That way we could keep the caching logic as an implementation detail, and it'd make indexPackageOwnerships more reusable/self-contained

pkg/cli/image_apk.go
```diff
@@ -0,0 +1,245 @@
...
+	owner pkg.Package
+	owned pkg.Package
+}
+
+func (r *syftResults) apks() []pkg.Package {
+	var apks []pkg.Package
+
+	pkgs := r.sbom.Artifacts.Packages.Sorted(pkg.ApkPkg)
+	for i := range pkgs {
+		p := pkgs[i]
```
pdeslaur, line 126: Is this better? More of a style question, but I wonder if it has different implications on the code below

```suggestion
	for i, p := range pkgs {
```

pdeslaur (commented):
Mostly bike shedding!

pkg/cli/image_apk.go
luhring: alsdkjalsdjalskdj — good catch!

pkg/cli/image_apk.go
luhring: I like this

pkg/cli/image_apk.go
luhring: Switched back to that and added a `nolint` — also happy to strike the lint rule globally if we want!

pkg/cli/image_apk.go
luhring: Switched back to that and added a `nolint` — also happy to strike the lint rule globally if we want!

pdeslaur (approved):

PR #571, opened 2024-01-24
Add `ruby` subcommand for working with ruby files

found-it:
Throughout my time working with ruby gems, I've noticed one thing. Typically I want to make one small change to a bunch of files, or do some small check across all github repositories. This usually ends up with me writing some temporary code to do the trick for me. In the process of getting ready to upgrade ruby packages to ruby 3.3 I actually put some of that temporary code into command form and now hopefully we have a useful `ruby` command.

The initial addition of this command is geared towards running some basic sniff tests across all ruby packages in the wolfi repo. Things like
* `required_ruby_version` explicitly rules out the upgraded ruby version
* The gem uses deprecated functions from stdlib (read from release notes)

These two goals allow us to take one step closer to the “factory” of images as we can automate the update of gems and run the wolfictl command to look for issues. The tooling should supports human encoded checks so upgrading the system ruby for all packages is as simple as reading the release notes and encoding checks then letting the tooling make decisions for us. It should allow us to provide a “green” signal in CI to upgrade packages to the next version while allowing humans to look at any packages flagged during the upgrade process.

I was able to run the command to get help focus the upgrade. It’s a bit rough still for processing the entire repo - I basically just dumped everything into a markdown file and inspected each flagged piece of information before making a final decision. In the future I expect it to work very similarly except for wolfictl to perform the upgrade and run these checks, then create a PR with any flagged information so a human can come inspect the PR and determine if there’s anything concerning. 53 of the packages made it through all checks without any flags meaning a little more than ⅓ of the packages could be auto-upgraded without human interaction.


pkg/cli/ruby.go
stormqueen1990, line 1: Random suggestion and feel free to disregard but could the subcommands be broken down into separate files?

pkg/ruby/check_upgrade.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	// search the gemspec for version constraints
+	return o.checkVersionConstraint(gemspecFile)
+}
+
+// findGemspec searches a given Github repository for a file named *.gemspec. It
+// searches for the file from the branch or tag specified in the melange yaml.
+func (o *Options) findGemspec(pkg *Package) (string, error) {
+	ctx := context.Background()
```
stormqueen1990, line 46: I'm wondering if the context should be coming from the invoker.

pkg/ruby/check_upgrade.go
```diff
@@ -0,0 +1,199 @@
...
+			break
+		}
+	}
+
+	// Could not find required_ruby_version in gemspec, assuming no constraint
+	if versionLine == "" {
+		return nil
+	}
+
+	pattern := `["']([^"']+)["']`
```
stormqueen1990, line 146: A few questions here:
* is the sole purpose of this regex to pull the version out of a constraint?
* is there any risk that the version string could be malformed (i.e. `"1.6.0'`)?
    * if malformed, do you still want to pull the version?

pkg/ruby/check_upgrade.go
```diff
@@ -0,0 +1,199 @@
...
+		// Need to construct a list of version constraints to pass
+		// to the ruby call. It should end up like this: ['>= 2.6', '< 4']
+		constraint := "["
+		for i, c := range versionConstraints {
+			constraint += fmt.Sprintf("'%s'", c)
+			if i != len(versionConstraints)-1 {
+				constraint += ", "
+			}
+		}
+		constraint += "]"
```
stormqueen1990, line 170: Can this be simplified with [`strings.Join()`](https://pkg.go.dev/strings#Join)?
I'm thinking something like:
```go
constraint := fmt.Sprintf("['%s']", strings.Join(versionConstraints, "', '"))
```

pkg/ruby/check_upgrade.go
```diff
@@ -0,0 +1,199 @@
...
+		}
+	} else { // If ruby doesn't exist on the system do the version comparison ourselves
+		// Terraform has the same version comparison operators as ruby so we
+		// just use the terraform library to do the comparison
+		v1, err := version.NewVersion(o.RubyUpdateVersion)
+		if err != nil {
+			return fmt.Errorf("converting ruby update version: %w", err)
+		}
+
+		constraints, err := version.NewConstraint(strings.Join(versionConstraints, ", "))
```
stormqueen1990, line 189: What's the difference between this and the preceding concatenation? Can you use the same result from before here?

pkg/ruby/discover_test.go
```diff
@@ -0,0 +1,144 @@
...
+		"ruby-33.2",
+		"ruby-2.2-dev",
+		"ruby-3.22-dev",
+		"ruby-33.2-dev",
+	}
+	for _, pkg := range shouldBeFalse {
+		testConfig.Environment.Contents.Packages[0] = pkg
+		assert.Falsef(t, o.isRubyPackage(testConfig), "%s should be false", pkg)
+	}
+}
```
stormqueen1990, line 47: This might make a good case to use table-based tests

pkg/gh/search.go
```diff
@@ -0,0 +1,45 @@
...
+			if rateLimited {
+				fmt.Printf("retrying after %v second delay due to rate limiting\n", delay.Seconds())
+				time.Sleep(delay)
+			} else {
+				return nil, githubErr
+			}
+		} else {
+			// if err is not rate limit, return err
+			return nil, err
+		}
```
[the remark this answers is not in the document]
found-it: Good note!

pkg/ruby/check_upgrade.go
found-it: Good questions! It is, I'd expect the version constraint to be properly formed since it's coming from the .gemspec which is valid ruby and would fail the gem build. Even if it was malformed I'd prefer to bias towards extracting as many version constraints as we can

stormqueen1990: Ah right, that makes sense. I forgot that would have to pass the scrutiny of build.

pkg/ruby/code_search.go
```diff
@@ -0,0 +1,179 @@
...
+type SearchResult struct {
+	Repository string
+	FilePath   string
+	URL        string
+	Fragments  []string
+}
+
+// CodeSearch is the main search function which uses Github Code Search to run
+// the specified query for a particular package.
+func (o *Options) CodeSearch(pkg *Package, query string) error {
```
joshrwolf, line 28:
```suggestion
func (o *Options) CodeSearch(ctx context.Context, pkg *Package, query string) error {
```

as mauren said, lets be good citizens and plumb context through 🙏

pkg/ruby/code_search.go
```diff
@@ -0,0 +1,179 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("marshaling json: %w", err)
+		}
+
+		_, err = cached.Write(jsonData)
+		if err != nil {
+			return nil, fmt.Errorf("failed to write cache file: %w", err)
+		}
+
+		_, err = cached.Seek(0, io.SeekStart)
```
joshrwolf, line 128: might be worth hoisting this leg out into a function so you can defer the closing without needing to rewind

pkg/ruby/check_upgrade.go
found-it: Love it!

pkg/ruby/check_upgrade.go
found-it: Unfortunately not, since one string is actually a valid array in ruby and the other is a function argument they are different

pkg/ruby/check_upgrade.go
found-it: Yep, fixing :)

pkg/ruby/discover_test.go
found-it: Good note! Converted

pkg/ruby/code_search.go
found-it: Plumbed ✅

pkg/ruby/code_search.go
found-it: Good note, did the hoisting. Happy to have any feedback :)

pkg/cli/ruby.go
found-it: I like it! Seems to match up with what some of the other subcommands are doing too.

pkg/ruby/code_search.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return fmt.Errorf("marshaling json: %w", err)
+	}
+
+	_, err = cached.Write(jsonData)
+	if err != nil {
+		return fmt.Errorf("failed to write cache file: %w", err)
+	}
+
+	defer cached.Close()
```
joshrwolf, line 175:
```suggestion
	cached, err := os.Create(cachedPath)
	if err != nil {
		return fmt.Errorf("failed to create cache file: %w", err)
	}
	defer cached.Close()


	// Convert the struct to JSON
	jsonData, err := json.Marshal(result)
	if err != nil {
		return fmt.Errorf("marshaling json: %w", err)
	}

	if _, err := cached.Write(jsonData); err != nil {
		return fmt.Errorf("failed to write cache file: %w", err)
	}
```

stormqueen1990: +1 to the suggestion

pkg/cli/ruby_check_upgrade.go
stormqueen1990, line 1: Thanks for splitting the files! 🙏

pkg/ruby/discover_test.go
stormqueen1990, line 1: These tests are just missing a `t.Run(name, func(t *testing.T){})` around each case to get handled as separate test cases.

pkg/ruby/code_search.go
```diff
@@ -0,0 +1,194 @@
...
+// CodeSearch is the main search function which uses Github Code Search to run
+// the specified query for a particular package.
+func (o *Options) CodeSearch(ctx context.Context, pkg *Package, query string) error {
+	results, err := o.runQuery(ctx, pkg, query)
+	if err != nil {
+		return fmt.Errorf("searching code: %w", err)
+	}
+
+	md := generateMarkdown(results, query)
+	fmt.Printf("%s", md)
```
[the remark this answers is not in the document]
found-it: Good catch :)

stormqueen1990 (approved):

joshrwolf (approved):

PR #573, opened 2024-01-24
ci: remove wolfictl text check

luhring:
There doesn't seem to be a clear sense of value for this check at this point, and the less time we spend waiting on CI the better.


jonjohnsonjr:
I think I prefer we keep this (to catch any changes in wolfictl that would break it) but prioritize fixing the data that is causing it to fail.

luhring:
SGTM!

PR #580, opened 2024-01-27
ci: bump golangci-lint to v1.55.2 and gitignore bin dir

Dentrax:


found-it (approved):

.github/workflows/golangci-lint.yaml
```diff
@@ -21,4 +21,4 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@3a919529898de77ec3da873e3063ca4b10e7f5cc # v3.7.0
         with:
-          version: v1.55.1
+          version: v1.55.2
```
luhring, line 24: I _think_ @cpanato told me once that if you drop the `.2` altogether it will automatically pick up new patch versions. Just something to consider! Not blocking.

cpanato: drop the patch version and use just the major.minor, then it will always get the latest for the minor

also pleas eupdate to `v1.57`

cpanato (changes requested):
see my comment

.github/workflows/golangci-lint.yaml
luhring: Fixed and force-pushed!

luhring:
Closing for now, looks like something is not working here

PR #590, opened 2024-01-29
wolfibot: go bump removals - use error messages that generate issues …

rawlingsj:
…rather than return errors that stop the automation

If we return an error then the bot will retry and then fail.  Instead we can return error messages that allow the bot to carry on to the next package update and will create issues for any error messages.


ajayk (approved):

luhring (approved):

hectorj2f:
Good catch! I've never guessed this behaviour.

PR #595, opened 2024-01-30
wolfibot: fix cloning repos where origin doesn't exist, it isn't need…

rawlingsj:
…ed for the clone to work


rawlingsj:
This will fix issues we just started seeing like https://github.com/wolfi-dev/os/issues/12057

jonjohnsonjr (approved):

PR #617, opened 2024-02-08
Advisory validation: start detecting duplicate IDs being used

luhring:
This will catch when both IDs and aliases are reused when they shouldn't be.

Example output:

```console
❌ advisory data is not valid.

basic validation failure(s):
    kube-state-metrics-2.6:
        advisories:
            CVE-2023-3978: advisory ID is not unique
            GHSA-2wrh-6pvc-2jm9: advisory alias is not unique
    metrics-server-fips:
        advisories:
            GHSA-45x7-px36-x8w8: advisory ID is not unique
```

### DIFF
--- a/pkg/configs/advisory/v2/document.go
+++ b/pkg/configs/advisory/v2/document.go
@@ -97,12 +97,32 @@ func (advs Advisories) Validate() error {
 		return fmt.Errorf("this file should not exist if there are no advisories recorded")
 	}
 
+	seenIDs := make(map[string]bool)
+	for _, adv := range advs {
+		if _, ok := seenIDs[adv.ID]; ok {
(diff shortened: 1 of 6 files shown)


pkg/configs/advisory/v2/document.go
```diff
@@ -97,13 +97,38 @@ func (advs Advisories) Validate() error {
 		return fmt.Errorf("this file should not exist if there are no advisories recorded")
 	}
 
+	seenIDs := make(map[string]bool)
```
pdeslaur, line 100: I love that we can easily improve after the big chunk of CI you added in 23d, this is awesome

pdeslaur (approved):

PR #629, opened 2024-02-14
Use sync.Cond for waiting on deps

jonjohnsonjr:
This should be a lot cheaper than just blocking on channels, since we know there's nothing left to do until the dep is finished.

I've also done a ReverseSorted on the dep graph so we start graph leaves first, while should help with making forward progress (and waiting on things that are more likely to have finished for status reporting).

### DIFF
--- a/pkg/cli/build.go
+++ b/pkg/cli/build.go
@@ -3,11 +3,12 @@ package cli
 import (
 	"context"
 	"fmt"
+	"log/slog"
 	"os"
 	"path/filepath"
 	"runtime"
 	"strings"
-	"time"
+	"sync"
 
 	"chainguard.dev/apko/pkg/build/types"
 	"chainguard.dev/melange/pkg/build"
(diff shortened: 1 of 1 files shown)


pkg/cli/build.go
```diff
@@ -153,52 +164,63 @@ type task struct {
...
+	done bool
+	cond *sync.Cond
 }
 
 func (t *task) start(ctx context.Context) {
-	log := clog.FromContext(ctx).With("pkg", t.pkg)
-	log.Infof("task %q waiting on %q", t.pkg, maps.Keys(t.deps))
+	defer func() {
+		t.cond.L.Lock()
+		clog.FromContext(ctx).Infof("finished %q, err=%v", t.pkg, t.err)
```
imjasonh, line 180:
```suggestion
		clog.FromContext(ctx).With("err", t.err).Infof("finished %q", t.pkg)
```

pkg/cli/build.go
```diff
@@ -153,52 +164,63 @@ type task struct {
...
 		return fmt.Errorf("failed to parse config: %w", err)
 	}
 
-	var bcs []*build.Build
 	for _, arch := range t.archs {
 		arch := types.ParseArchitecture(arch).ToAPK()
-		log := clog.New(log.Handler()).With("arch", arch)
+
+		// TODO: Handle these logs in an interesting way instead of discarding them.
+		log := clog.New(discard).With("arch", arch)
```
imjasonh, line 222: Why not `log.Handler()` like we had before?

PR #649, opened 2024-02-22
build: Emulate SOURCE_DATE_EPOCH from Makefile

jonjohnsonjr:


.golangci.yaml
```diff
@@ -26,7 +26,6 @@ linters:
     - dupl
     - durationcheck
     - errcheck
-    - goconst
```
jonjohnsonjr, line 29: I'm not making `.yaml` a constant, this lint is bad.

imjasonh (approved):

PR #677, opened 2024-03-06
wolfictl pmub: just a reverse of wolfictl bump

ajayk:


jonjohnsonjr:
What if this was just a flag on `bump`?

ajayk:
@jonjohnsonjr PTAL

pkg/cli/bump.go
```diff
@@ -120,6 +129,7 @@ modifying anything in the filesystem.
 	cmd.Flags().BoolVar(&opts.epoch, "epoch", true, "bump the package epoch")
 	cmd.Flags().BoolVar(&opts.dryRun, "dry-run", false, "don't change anything, just print what would be done")
 	cmd.Flags().StringVar(&opts.repoDir, "repo", ".", "path to the wolfi/os repository")
+	cmd.Flags().BoolVar(&opts.increment, "increment", true, "increments/decrements the package epoch")
```
jonjohnsonjr, line 132: Would we ever want to bump by anything other than `-1` and `+1`? Would it make sense to have this take an int flag that defaults to `1`?

E.g. to bump down, do this:
```
wolfictl bump --by -1 foo.yaml
```

If not, then I think I'd prefer this be a negative flag (`--decrement` or `--substract` or `--minus` or `--undo`) that defaults to `false` because it's easier to type:

```
wolfictl bump --undo
```

Than it is to type:

```
wolfictl bump --increment=false
```

But also because I just learned that this would be an error:

```
wolfictl bump --increment false
```

(See https://github.com/wolfi-dev/os/commit/bfe82af1d3d9c22c3a003520e1f961e0d7a12f68 🤦)

PR #692, opened 2024-03-12
Add filter-prefix support for 'wolfictl check update'

smoser:
This relies on
https://github.com/chainguard-dev/melange/pull/1083

And allows adding an "include prefix" for update checks.


smoser:
going to close this for now, to be revived if needed.

PR #705, opened 2024-03-25
Revert "build: Un-group by arch"

imjasonh:
Reverts wolfi-dev/wolfictl#704


xnox (approved):
Revert is correct diff, returning to state before previous merge of this feature.

PR #708, opened 2024-03-26
Revert my life's work

jonjohnsonjr:


ajayk (approved):

xnox (approved):
This brings us back to current production level wolfictl sdk version. W.r.t. the build command, but allows upgrades of the sdk container to newer things (like the OSV and like libxcrypt split).

PR #731, opened 2024-04-01
build: Consider subpackages for build order

jonjohnsonjr:
Previously, we would just filter out any non-main packages in our build graph. Unfortunately, that filtered out any edges to subpackages that were important for build ordering.

As one real example, consider this dependency:

  spdlog -> fmt-dev

Since spdlog depends on fmt-dev, we need to build fmt-dev before we can build spdlog.

We can't build fmt-dev directly, since it's a subpackage of fmt, so we need to build fmt before we build spdlog.

Before this change, that edge would just get dropped and we would try to build spdlog and fmt at the same time. The spdlog build would fail because fmt-dev would not exist until after fmt completes.

With this change, dag.Graph.Targets() essentially flattens any subpackage nodes in the graph so that they are equivalent to the main package.

So after calling Targets(), that edge looks like this:

  spdlog -> fmt

Which preserves the necessary build order rather than ignoring it.

The reason this hasn't been a problem in the past is that we usually don't introduce two packages at the same time that have a dependency between them, so builds justs fall back to the version that already exists.

### DIFF
--- a/pkg/cli/build.go
+++ b/pkg/cli/build.go
@@ -148,9 +148,9 @@ func walkConfigs(ctx context.Context, cfg *global) (*configStuff, error) {
 	}
 
 	// Only return main packages (configs) because we can't build just subpackages.
-	g, err = g.Filter(dag.OnlyMainPackages(pkgs))
+	g, err = g.Targets()
 	if err != nil {
-		return nil, err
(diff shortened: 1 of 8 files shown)


imjasonh (approved):
Thank you thank you thank you!

PR #733, opened 2024-04-02
add extra packages distro

cpanato:
in preparation for the new advisories repo for extra packages


luhring (approved):
Looks great! And hopefully pretty painless 😃 

I've wondered if we should have more tests here (even before this PR), but most of what could go wrong is dependent on the repos and other references defined for the distro being set up correctly, i.e. external to `wolfictl`.

Thanks for doing this!

PR #742, opened 2024-04-04
hoist some methods into functions to reuse

joshrwolf:
renames some things and refactors some methods for a follow up `wolfictl test` PR


joshrwolf:
this makes more sense with everything, so closing in favor of #743

PR #631, opened 2024-02-14
feat(adv): interactive guide for advisory entry

luhring:
This introduces a new command: `wolfictl adv guide`. This command is a new approach to responding to vulnerabilities in our distro packages by entering the appropriate advisory data.

This is a work in progress!

image

PR #759, opened 2024-04-15
advisory guide: enabling advisory data entry by interview

luhring:
This PR is a (substantial) follow-up to https://github.com/wolfi-dev/wolfictl/pull/631, which introduced a partially implemented `wolfictl adv guide` command.

To recap, the purpose of this command is to rethink the user experience of entering advisory data for a package.

This PR introduces several notable improvements to the original state:

1. **Short-lived advisory data sessions** — This command is the first to try a new approach, where `wolfictl` no longer expects you to have cloned any advisories repos in advance. Instead, when you run the command, it grabs the latest data from GitHub, makes updates to that data on your behalf locally, and then lets you open a PR with your changes — all without having had to interact with YAML. This is driven by a new internal `advisory.DataSession` object which manages the session for you.

2. **"Interviews"** — A concept where we can ask a series of questions to the user, where each answer might result in an update to state (e.g. the details of a new advisory entry). The questions and answers compose a directed labeled graph of sorts, which themselves are decoupled from the terminal UI altogether, so we can iterate on the questions and answers quickly.

3. **"Build groups"** — This lets `wolfictl` now operate on a Melange-produced origin APK with its subpackage APKs as a contextualized set.

### What's missing

To be useful, we need to iterate on the **interview questions** and answers to the point where they collectively cover an acceptable portion of cases that engineers encounter when triaging vulnerabilities in newly built APKs.

I've also been making note of useful **UX polish** opportunities, but I've been deferring them for now to facilitate faster iteration. But as users begin trying this out, it'd be great to respond to their feedback and minimize the friction associated with this command.

This also needs to be further tested with **workstations**, to ensure auth and the rest of the workflow behaves as intended.

### DIFF
--- a/pkg/advisory/data_session.go
+++ b/pkg/advisory/data_session.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"slices"
 	"strings"
 
 	"github.com/go-git/go-git/v5"
@@ -180,8 +181,10 @@ func (ds DataSession) Push(ctx context.Context) error {
 
 // OpenPullRequest opens a pull request for the changes made during the session.
 func (ds DataSession) OpenPullRequest(ctx context.Context) (*PullRequest, error) {
+	slices.Sort(ds.modifiedPackages)
+	compact := slices.Compact(ds.modifiedPackages)
 	newPullRequest := github.NewPullRequest{
-		Title:               github.String(fmt.Sprintf("Add advisory data for %s", strings.Join(ds.modifiedPackages, ", "))),
+		Title:               github.String(fmt.Sprintf("Add advisory data for %s", strings.Join(compact, ", "))),
 		Body:                github.String(pullRequestBody),
 		Head:                github.String(ds.workingBranch),
 		Base:                github.String("main"),
--- a/pkg/advisory/data_session.go
+++ b/pkg/advisory/data_session.go
@@ -1,0 +1,237 @@
+package advisory
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/google/go-github/v58/github"
+	"github.com/google/uuid"
+	"github.com/wolfi-dev/wolfictl/pkg/configs"
+	v2 "github.com/wolfi-dev/wolfictl/pkg/configs/advisory/v2"
+	rwos "github.com/wolfi-dev/wolfictl/pkg/configs/rwfs/os"
+	"github.com/wolfi-dev/wolfictl/pkg/distro"
+	wgit "github.com/wolfi-dev/wolfictl/pkg/git"
+)
+
+type DataSession struct {
+	tempDir          string
+	repo             *git.Repository
+	workingBranch    string
+	distro           distro.Distro
+	index            *configs.Index[v2.Document]
+	githubClient     *github.Client
+	modified         bool
+	modifiedPackages []string
+}
+
+type DataSessionOptions struct {
+	Distro       distro.Distro
+	GitHubClient *github.Client
+}
+
+// NewDataSession initializes a new advisory data session for the specified
+// distro and returns a reference to the session. This call will retrieve the
+// data and manage it in a local temp directory until the session is closed. The
+// session should be closed by calling Close() when it is no longer needed.
+func NewDataSession(ctx context.Context, opts DataSessionOptions) (*DataSession, error) {
+	// create temp directory
+	tempDir, err := os.MkdirTemp("", "wolfictl-advisory-data-session-*")
+	if err != nil {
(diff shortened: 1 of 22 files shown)


cpanato (approved):
i did a review this morning and was waiting to clarify a question with @luhring all good now

thanks for this

pkg/advisory/data_session.go
```diff
@@ -0,0 +1,240 @@
...
+
+// Modified returns true if any changes have been made to the advisory data
+// during the session.
+func (ds DataSession) Modified() bool {
+	return ds.modified
+}
+
+// Push pushes the changes made during the session to the remote advisories
+// repository.
+func (ds DataSession) Push(ctx context.Context) error {
```
hectorj2f, line 170: It doesn't seem we are using `Push` and OpenPullRequest anywhere else.

hectorj2f (commented):
Could we add some tests to validate these changes ?

luhring:
> Could we add some tests to validate these changes ?

@hectorj2f Sure thing — what did you have in mind? The vast majority of this PR is command logic and components, which are difficult to test directly, but I'm happy to try something. We could add tests for the Data Session pretty easily, though.

pkg/advisory/data_session.go
luhring: What do you mean exactly?

https://github.com/luhring/wolfictl/blob/b553e73abef34930ebb96893c194e9f44af0a0b7/pkg/cli/advisory_guide.go#L352

https://github.com/luhring/wolfictl/blob/b553e73abef34930ebb96893c194e9f44af0a0b7/pkg/cli/advisory_guide.go#L357

pkg/distro/detectv2.go
```diff
@@ -0,0 +1,86 @@
...
+	"os"
+	"slices"
+
+	"github.com/go-git/go-git/v5"
+)
+
+// DetectV2 tries to detect which distro the user wants to operate on by seeing
+// if the current working directory is the distro's packages repo. No
+// advisory-related repositories are detected.
+func DetectV2() (Distro, error) {
```
hectorj2f, line 14: **Suggestion**: I don't have strong feelings about this functionality. I am just a bit concerned of the value of adding this detection versus explicitly requesting the distro to users.

pkg/advisory/data_session.go
hectorj2f: Okay, I see them now. It was github failed to me. 👍🏻

pkg/cli/advisory_guide.go
```diff
@@ -295,16 +511,100 @@ var (
...
+
+					req.Event = e
+					return req, &questionWhyFalsePositive
+				},
+			},
+		},
+	}
+
+	questionWhyFalsePositive = questions.Question[advisory.Request]{
+		Text: "Why is this a false positive?",
```
hectorj2f, line 554: **Question**: where should it go ? `I ran `-govulncheck` and tells me this is a false positive...` Do we need to add this case ?

pkg/distro/detectv2.go
luhring: What do you mean by "explicitly requesting the distro to users"?

Detecting the distro this way was a suggestion from the delivery squad, and it's _less implicit_ than the `Detect()` function we already use elsewhere in the advisory commands, since it no longer detects your advisory repos.

pkg/cli/advisory_guide.go
luhring: Yes, the questions aren't completed yet, but we can iterate on those next...

> What's missing
> 
> To be useful, we need to iterate on the interview questions and answers to the point where they collectively cover an acceptable portion of cases that engineers encounter when triaging vulnerabilities in newly built APKs.

hectorj2f (commented):
> @hectorj2f Sure thing — what did you have in mind? The vast majority of this PR is command logic and components, which are difficult to test directly, but I'm happy to try something. We could add tests for the Data Session pretty easily, though.

Generally I expected to see a higher test coverage. I understand certain functions are difficult to test. But I feel we could add tests to certain functions or re-organize the code in these functions to be tested.

pkg/cli/advisory_guide.go
hectorj2f: sgtm

pkg/distro/detectv2.go
hectorj2f: Good. I was missing that context 👍🏻 . That helped to alleviate my concerns 😃 .

luhring:
> > @hectorj2f Sure thing — what did you have in mind? The vast majority of this PR is command logic and components, which are difficult to test directly, but I'm happy to try something. We could add tests for the Data Session pretty easily, though.
> 
> Generally I expected to see a higher test coverage. I understand certain functions are difficult to test. But I feel we could add tests to certain functions or re-organize the code in these functions to be tested.

I'm open to suggestions!

luhring:
Merging for now, but suggestions for improvement are welcome and we can follow-up with as many PRs as we need 🙇

PR #772, opened 2024-04-22
check: increase bincapz file threshold from 1 to 3

tstromberg:
This is to lower the amount of bincapz change noise we see.

Alternatively, we should set a default flag value for GitHub actions, but then we run into the messy issue of "wolfictl check" behaving different than our GHA checks.


tstromberg:
cc @mattmoor @rawlingsj

k4leung4 (approved):

PR #777, opened 2024-04-24
fix: compare package version epoch as numerics

hectorj2f:
This is related to the issue reported on slack [here](https://chainguard-dev.slack.com/archives/C05GYUBM07Q/p1713943656535999) where a list of versions as such `[0.1.0-r12, 0.1.0-r9, 0.1.0-r1]` always returned `0.1.0-r9` as the latest.

### DIFF
--- a/pkg/versions/versions.go
+++ b/pkg/versions/versions.go
@@ -1,10 +1,13 @@
 package versions
 
 import (
+	"strconv"
 	"strings"
 
 	"github.com/hashicorp/go-version"
 )
+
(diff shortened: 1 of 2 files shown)


luhring (commented):
Thanks! Could we just add a few test cases to validate this? Maybe in `TestSort_ByLatestStrings`?

luhring (approved):
Nice!

cpanato:
ok we need to release this and bump the dependencies on the infra-images side

PR #789, opened 2024-04-25
fix: get latest package version

hectorj2f:
Fixing the code fetching the latest package version in a different place. There is already a test validating this function.

### DIFF
--- a/pkg/apk/apk.go
+++ b/pkg/apk/apk.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"log"
 	"net/http"
+	"sort"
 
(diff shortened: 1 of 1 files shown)


cpanato (approved):

luhring:
@hectorj2f Quick question — 

>There is already a test validating this function.

If there's already a test to catch this, are you saying it was failing? If it was passing then I think we need to augment the test coverage.

hectorj2f:
@luhring Sounds fair! [👍🏻 ](https://github.com/wolfi-dev/wolfictl/pull/790)

PR #807, opened 2024-05-06
new triage framework and automated triaging!

luhring:
### Design 🗺️ 

This introduces a **new triaging framework** centered on this new interface:

```go
type Triager interface {
	// Triage takes a scan result and returns a triage conclusion.
	Triage(ctx context.Context, vulnFindings scan.VulnFindings) (*advisory.Request, error)
}
```

The idea is that we arrange findings for a single vulnerability in an APK package as a `scan.VulnFindings`, such that triager implementations can take this and return their conclusion on whether the vulnerability truly affects the APK package. The result is an `advisory.Request` — i.e. new advisory data all ready to be applied to our advisory data set.

The consumer can wire up as many triagers as needed. Each triager can bring along its own state or network clients as necessary. So while some triagers may only need access to the APK data, others might want to access other local information or even do online lookups.

### Features 🚀 

Automated triaging! 🎉 

This PR adds **two concrete implementations** of the Triager interface, **which have already been massively helpful at dealing with vulnerabilities** (these dealt with ~130 vulnerabilities in ~10 minutes net runtime):

1. "govulncheck" — Based on the _tool_ govulncheck, this implementation cross-references the input findings with an analysis of symbols found in Go binaries to decide if an otherwise present vulnerability has no impact on the APK.
2. "gogitversion" — This clones git repositories for affected Go modules to do version-control-aware analysis of whether remediating commits have been incorporated into the version of the Go module used in a binary in an APK package. The implementation maintains a local cache of git repositories to drastically speed up triaging.

Orthogonal to these features, this PR also was designed for incorporating into future **services**, and therefore has been instrumented with lots of **logging** and **tracing**. 📊 

### Caveats ☠️ 

Since this PR was written to illustrate a concept, with more work originally intended to come later, there are several aspects that would need further work before considering this logic ready-to-use in production.

1. There are no tests — and this is the kind of logic that needs THOROUGH testing. We cannot afford inaccurate advisory data to be automatically added to our data set without proper verification of the validity of triaging assessments.
2. There are many spots in the execution when context cancellation should be considered due to the long-running nature of several pieces in the logic flow.
3. We need a more robust strategy for mapping fixed version data to tags in a source code repo (for "gogitversion"); so far I've been manually adjusting code to expect a leading `v` prefix and to not expect one.
4. The "gogitversion" implementation can likely be generalized to handle cases beyond Go.
5. The commit-to-later-tag version check needs to be taken with a grain of salt, since we're not checking all affected version ranges, only a simple comparison to the "fixed version" reported by Grype.
6. This code doesn't understand the notion of APK "build groups" yet (i.e. an origin packages and its subpackages all produced in the same Melange build). This matters! Because we should not draw a conclusion about a vulnerability affecting an APK package without assessing the impact on the entire build group, since the advisory data is used to apply to the entire build group.
7. Same thing as above, but for multiple architectures instead of multiple packages within a build group.
8. More care should be taken on understanding the latest advisory data for a given vulnerability before proceeding to triage. For example, if we've already drawn our conclusion about a vulnerability (say as a TP or FP), we probably don't want to automatically append an updated triage assessment after that, in contexts when we're not certain that the automated triaging has as much context as went into the previous triage assessment.
9. We should use a better source of truth for fixed version data than the Grype database. Perhaps a local OSV database.

### DIFF
--- a/pkg/cli/advisory_triage.go
+++ b/pkg/cli/advisory_triage.go
@@ -1,0 +1,283 @@
+package cli
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"path/filepath"
+
+	"github.com/adrg/xdg"
+	"github.com/anchore/grype/grype/db"
+	"github.com/anchore/grype/grype/db/v5/store"
+	"github.com/chainguard-dev/clog"
+	"github.com/chainguard-dev/go-apk/pkg/expandapk"
+	"github.com/chainguard-dev/yam/pkg/yam/formatted"
+	"github.com/spf13/cobra"
+	"github.com/wolfi-dev/wolfictl/pkg/advisory"
+	"github.com/wolfi-dev/wolfictl/pkg/cli/styles"
+	v2 "github.com/wolfi-dev/wolfictl/pkg/configs/advisory/v2"
+	rwos "github.com/wolfi-dev/wolfictl/pkg/configs/rwfs/os"
+	"github.com/wolfi-dev/wolfictl/pkg/scan"
+	"github.com/wolfi-dev/wolfictl/pkg/scan/target"
+	"github.com/wolfi-dev/wolfictl/pkg/scan/triage"
+	"github.com/wolfi-dev/wolfictl/pkg/scan/triage/gogitversion"
+	"github.com/wolfi-dev/wolfictl/pkg/scan/triage/govulncheck"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/exporters/stdout/stdouttrace"
+	"go.opentelemetry.io/otel/sdk/trace"
+	"gopkg.in/yaml.v3"
+)
+
+//nolint:gocyclo // we should refactor this function when the design is more solid
+func cmdAdvisoryTriage() *cobra.Command {
+	p := &triageParams{}
+	cmd := &cobra.Command{
+		Use:   "triage",
+		Short: "Triage vulnerabilities and record analysis in advisory data",
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			ctx := cmd.Context()
+
+			logger := clog.NewLogger(getLogger(p.verbosity))
+			ctx = clog.WithLogger(ctx, logger)
+
+			if p.traceFile != "" {
+				w, err := os.Create(p.traceFile)
+				if err != nil {
+					return fmt.Errorf("creating trace file: %w", err)
+				}
+				defer w.Close()
+				exporter, err := stdouttrace.New(stdouttrace.WithWriter(w))
+				if err != nil {
+					return fmt.Errorf("creating stdout exporter: %w", err)
+				}
+				tp := trace.NewTracerProvider(trace.WithBatcher(exporter))
+				otel.SetTracerProvider(tp)
+
+				defer func() {
+					if err := tp.Shutdown(context.WithoutCancel(ctx)); err != nil {
+						clog.FromContext(ctx).Errorf("shutting down trace provider: %v", err)
+					}
+				}()
+
+				tctx, span := otel.Tracer("wolfictl").Start(ctx, "advisory triage")
+				defer span.End()
+				ctx = tctx
+
+				logger.Info("tracing enabled", "traceFile", p.traceFile)
+			}
+
+			// TODO: Support triaging multiple packages at once
+			if p.packageName == "" {
+				return fmt.Errorf("--%s is required", flagNamePackage)
+			}
+
+			if p.advisoriesDir == "" {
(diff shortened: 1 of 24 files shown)


pkg/internal/gomod/repo.go
```diff
@@ -0,0 +1,39 @@
...
+	"fmt"
+	"net/http"
+	"regexp"
+	"strings"
+
+	"go.opentelemetry.io/otel"
+)
+
+// Repo returns the source code repository URL for the given module import path.
+func Repo(ctx context.Context, importpath string) (string, error) {
```
hectorj2f, line 14: We could add some tests about this function. I added some on the infra-images code.

pkg/internal/gomod/pseudoversion.go
```diff
@@ -0,0 +1,14 @@
...
+import "regexp"
+
+var (
+	pseudoVersionPattern = `^v\d+\.\d+\.\d+-\d{14}-[0-9a-f]{12}$`
+	pseudoVersionRegexp  = regexp.MustCompile(pseudoVersionPattern)
+)
+
+// IsPseudoVersion reports whether the given version is a Go pseudo-version
+// (https://go.dev/ref/mod#pseudo-versions).
+func IsPseudoVersion(version string) bool {
```
hectorj2f, line 12: Also let's add some tests here.

pkg/internal/gomod/repo.go
```diff
@@ -0,0 +1,39 @@
...
+	_, span := otel.Tracer("wolfictl").Start(ctx, fmt.Sprintf("resolving git URL for go module %s", importpath))
+	defer span.End()
+
+	if strings.HasPrefix(importpath, "github.com/") {
```
hectorj2f, line 18: This only matches github.com but we have certain projects using gitlab.com. Let's add both use cases.

pkg/cli/advisory_triage.go
```diff
@@ -0,0 +1,283 @@
...
+			defer os.RemoveAll(expandedAPKTempDir)
+			apkExpanded, err := expandapk.ExpandApk(ctx, apk, expandedAPKTempDir)
+			if err != nil {
+				return fmt.Errorf("expanding APK: %w", err)
+			}
+
+			// For now, the implementation peeks into the Grype database. But it'd be good
+			// to replace this with a more upstream way to get the vulnerability data, such
+			// as local OSV data stores.
+			grypeStore, err := store.New(filepath.Join(scan.GrypeDBDir(), "5", "vulnerability.db"), false)
```
hectorj2f, line 122: This sounds like a `TODO` here. Couldn't we use `LoadVulnerabilityDB` as done for scanning or similar.

luhring: `LoadVulnerabilityDB` does a lot more things than we need here. It probably shouldn't, but it does for now.

This implementation boils down what we need to a minimum. It's also _hopefully_ short-lived — I'd love to get the local OSV DB thing working at some point.

pkg/cli/advisory_triage.go
```diff
@@ -0,0 +1,283 @@
...
+	distroDir            string
+	upstreamRepoCacheDir string
+	packageName          string
+	traceFile            string
+	dryRun               bool
+}
+
+func (p *triageParams) addFlagsTo(cmd *cobra.Command) {
+	addVerboseFlag(&p.verbosity, cmd)
+	addBuiltPackagesDirFlag(&p.builtPackagesDir, cmd)
```
hectorj2f, line 271: What about downloading the package ?

hectorj2f:
@luhring I have tried the cli to generate some advisory data but i didn't get anything and the command exited successfully.

```
wolfictl adv triage -a ../advisories -p ingress-nginx-controller --built-packages-dir ~/os/packages/aarch64
 ```

File: pkg/scan/triage/gogitversion/triager.go [hunk 6]
```diff
@@ -0,0 +1,714 @@
...
+	// module findings.
+	if adjuster, ok := versionTagAdjustersByGoModule[moduleName]; ok {
+		fixedVersion = adjuster(fixedVersion)
+	}
+
+	// Grype sometimes trims the "v" prefix in the fixed version, which can be fixed
+	// like this:
+
+	//nolint:gocritic // intentionally commented out
+	// if !strings.HasPrefix(fixedVersion, "v") {
```
hectorj2f, line 398: I faced issues with this. It didn't find the tag so it reached no conclusions.

luhring:
Yeah @hectorj2f this isn't meant to be used yet 😄 still a draft PR — I just wanted to show the general idea and the progress toward that idea so far

luhring:
Note: The team has decided not to pursue merging this PR. We'll leave it here for a short time to serve as a reference for other work.

hectorj2f:
Thanks @luhring 👍🏻 !

PR #808, opened 2024-05-06
if the package only exist for x86 dont fail and continue the scanning

cpanato:
before
```
 wolfictl scan -r chromium                                                                                                                                                                                                                                                                                                                                                                                                                                                                               📡 Finding remote packages
2024/05/06 15:52:37 INFO error during command execution: failed to resolve input "chromium" for remote scanning: no Wolfi package found with name "chromium" in arch "aarch64"
```


after

```
$ wolfictl scan -r chromium
📡 Finding remote packages
🔎 Scanning "/var/folders/kl/q9mydw095ln5s7wj971qcrx40000gn/T/x86_64-chromium-124.0.6367.91-r0-27198284.apk"
└── 📄 /.PKGINFO
        📦 chromium 124.0.6367.91-r0 (apk)
            Unknown CVE-2024-4331
            Unknown CVE-2024-4368

```

### DIFF
--- a/pkg/cli/scan.go
+++ b/pkg/cli/scan.go
@@ -626,7 +626,13 @@ func resolveInputForRemoteTarget(ctx context.Context, input string) (downloadedA
 		})
 
 		if len(nameMatches) == 0 {
-			return nil, nil, fmt.Errorf("no Wolfi package found with name %q in arch %q", input, arch)
+
+			if arch != "aarch64" {
+				return nil, nil, fmt.Errorf("no Wolfi package found with name %q in arch %q", input, arch)
+			}
+
+			logger.Warnf("no Wolfi package found with name %q in arch %q but will continue, it might not have a package built for this arch.", input, arch)
+			continue
 		}
 
 		vers := lo.Map(nameMatches, func(pkg *apk.Package, _ int) string {


pkg/cli/scan.go
```diff
@@ -626,7 +626,13 @@ func resolveInputForRemoteTarget(ctx context.Context, input string) (downloadedA
...
 
 		if len(nameMatches) == 0 {
-			return nil, nil, fmt.Errorf("no Wolfi package found with name %q in arch %q", input, arch)
+
+			if arch != "aarch64" {
+				return nil, nil, fmt.Errorf("no Wolfi package found with name %q in arch %q", input, arch)
+			}
+
+			logger.Warnf("no Wolfi package found with name %q in arch %q but will continue, it might not have a package built for this arch.", input, arch)
+			continue
```
luhring, line 635: I think this idea makes sense, and logging a warning in this case is a great call.

Could we make it architecture agnostic? If any arch is missing, but we still find at least one, proceed with scanning but log a warning. If no arches found, error out.

cpanato: done

PTAL

pkg/cli/scan.go
```diff
@@ -614,6 +614,7 @@ func resolveInputFileFromArg(inputFilePath string) (*os.File, error) {
 func resolveInputForRemoteTarget(ctx context.Context, input string) (downloadedAPKFilePaths []string, cleanup func() error, err error) {
 	logger := clog.FromContext(ctx)
 
+	noPackagesFound := 0
 	for _, arch := range []string{"x86_64", "aarch64"} {
```
luhring, line 618: This looks like it will work today, I just have on thought to make it less likely to break in the future — could we invert this and say `archesFound := 0`, and then increment it when we **_do find_** a package for that architecture. Down below, we could have the error condition be `if archestFound == 0` instead of `if noPackagesFound == 2`. WDYT? I don't feel strongly, it just seems a bit more straightforward to my brain

cpanato: sound great that!
ptal

luhring (approved):

PR #815, opened 2024-05-07
Update melange dependency for 2 'provides' changes.

smoser:
Only 2 changes being pulled in here.
1. Only consider commands in a PATH dir from generateCmdProviders https://github.com/chainguard-dev/melange/pull/1164
2. Allow symlinks to provide cmd: https://github.com/chainguard-dev/melange/pull/1188

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.22.0
 
 require (
 	chainguard.dev/apko v0.14.1-0.20240308000904-c510767a86aa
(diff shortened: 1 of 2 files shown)


smoser:
I'm under the impression that this is required to get the melange changes into c-i (so that it actually changes builds).  If that is not the case, then this PR is of no real value.

imjasonh (approved):

PR #830, opened 2024-05-13
Improve 'wolfictl check update' usage to include yaml.

smoser:
The previous usage statement just showed:

    wolfictl check update [flags]

Which did not make it obvious that you could pass a melange .yaml file in.

### DIFF
--- a/pkg/cli/check_update.go
+++ b/pkg/cli/check_update.go
@@ -14,7 +14,7 @@ func CheckUpdate() *cobra.Command {
 	}
 
 	cmd := &cobra.Command{
-		Use:               "update",
+		Use:               "update [config[.yaml]...]",
 		DisableAutoGenTag: true,
 		SilenceUsage:      true,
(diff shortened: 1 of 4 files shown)


pkg/cli/check_update.go
```diff
@@ -14,7 +14,7 @@ func CheckUpdate() *cobra.Command {
 	}
 
 	cmd := &cobra.Command{
-		Use:               "update",
+		Use:               "update [flags] [config[.yaml]...]",
```
luhring, line 17:
```suggestion
		Use:               "update [config[.yaml]...]",
```

nit: cobra adds the `[flags]` bit automatically, we can just add the config file part.

luhring (approved):
I love this. 😍 

I wish we used the `Use:` field in cobra correctly like this more of the time!

luhring (approved):
Thanks for making the code better!

PR #835, opened 2024-05-14
introduce CGA ids to the advisories

cpanato:
- introduce CGA ids to the advisories

docs: https://docs.google.com/document/d/1cLTpQp-dknQ2HU3EKmAQMPeOGR9MBQ_IfQ2T7_ZUb40/edit

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,7 +1,7 @@
 ---
 run:
   concurrency: 6
-  deadline: 5m
+  timeout: 10m
 issues:
   exclude-rules:
     # counterfeiter fakes are usually named 'fake_<something>.go'
@@ -9,6 +9,9 @@ issues:
       linters:
         - gocritic
         - golint
+        - dupl
+    - path: .*_test.go
+      linters:
         - dupl
 
   # Maximum issues count per one linter. Set to 0 to disable. Default is 50.
@@ -99,22 +102,8 @@ linters-settings:
   gocritic:
     enabled-checks:
       # Diagnostic
-      - appendAssign
-      - argOrder
-      - badCond
-      - caseOrder
-      - codegenComment
       - commentedOutCode
-      - deprecatedComment
-      - dupArg
-      - dupBranchBody
-      - dupCase
-      - dupSubExpr
-      - exitAfterDefer
-      - flagDeref
-      - flagName
       - nilValReturn
-      - offBy1
       - sloppyReassign
       - weakCond
       - octalLiteral
@@ -128,31 +117,16 @@ linters-settings:
       - rangeValCopy
 
       # Style
-      - assignOp
       - boolExprSimplify
-      - captLocal
-      - commentFormatting
       - commentedOutImport
-      - defaultCaseOrder
       - docStub
-      - elseif
       - emptyFallthrough
       - emptyStringTest
       - hexLiteral
       - methodExprCall
-      - regexpMust
-      - singleCaseSwitch
-      - sloppyLen
       - stringXbytes
-      - switchTrue
       - typeAssertChain
-      - typeSwitchVar
-      - underef
       - unlabelStmt
-      - unlambda
-      - unslice
-      - valSwap
-      - wrapperFunc
       - yodaStyleExpr
       # - ifElseChain
 
@@ -169,9 +143,6 @@ linters-settings:
   nolintlint:
     # Enable to ensure that nolint directives are all used. Default is true.
     allow-unused: false
-    # Disable to ensure that nolint directives don't have a leading space. Default is true.
-    # TODO(lint): Enforce machine-readable `nolint` directives
-    allow-leading-space: true
     # Exclude following linters from requiring an explanation.  Default is [].
     allow-no-explanation: []
     # Enable to require an explanation of nonzero length after each nolint directive. Default is false.
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -99,6 +99,12 @@ linters-settings:
   errcheck:
     check-type-assertions: true
     check-blank: true
+  gosec:
+    # To specify the configuration of rules.
+    config:
+      # Maximum allowed permissions mode for os.WriteFile
+      # Default: "0600"
+      G306: "0644"
   gocritic:
     enabled-checks:
       # Diagnostic
--- a/pkg/advisory/cga_id.go
+++ b/pkg/advisory/cga_id.go
@@ -1,0 +1,42 @@
+package advisory
+
+import (
+	"crypto/rand"
+	"fmt"
+	"math/big"
+)
+
+func GenerateCGAID() (string, error) {
+	// Allowed characters [23456789cfghjmpqrvwx]
+	allowedChars := "23456789cfghjmpqrvwx"
+
+	// Length of the required UUID parts
+	partLength := 4
+	totalLength := partLength * 3
+
+	// Function to get a random character from allowedChars
+	getRandomChar := func() (byte, error) {
+		max := big.NewInt(int64(len(allowedChars)))
+		n, err := rand.Int(rand.Reader, max)
+		if err != nil {
+			return 0, err
+		}
+		return allowedChars[n.Int64()], nil
+	}
+
+	// Generate random characters
+	randomChars := make([]byte, totalLength)
+	for i := 0; i < totalLength; i++ {
+		char, err := getRandomChar()
+		if err != nil {
+			return "", err
+		}
+
+		randomChars[i] = char
+	}
+
+	// Format the custom UUID to match CGA(-[23456789cfghjmpqrvwx]{4}){3}
+	formattedUUID := fmt.Sprintf("CGA-%s-%s-%s", randomChars[0:4], randomChars[4:8], randomChars[8:12])
+
+	return formattedUUID, nil
+}
--- a/pkg/advisory/cga_id.go
+++ b/pkg/advisory/cga_id.go
@@ -35,7 +35,7 @@ func GenerateCGAID() (string, error) {
 		randomChars[i] = char
 	}
 
-	// Format the custom UUID to match CGA(-[23456789cfghjmpqrvwx]{4}){3}
+	// Format the custom UID to match CGA(-[23456789cfghjmpqrvwx]{4}){3}
 	formattedUUID := fmt.Sprintf("CGA-%s-%s-%s", randomChars[0:4], randomChars[4:8], randomChars[8:12])
 
 	return formattedUUID, nil
--- a/pkg/advisory/cga_id.go
+++ b/pkg/advisory/cga_id.go
@@ -6,7 +6,7 @@ import (
 	"time"
 )
 
-func GenerateCGAID() (string, error) {
+func GenerateCGAIDWithSeed(seed int64) (string, error) {
 	// Allowed characters [23456789cfghjmpqrvwx]
 	allowedChars := "23456789cfghjmpqrvwx"
 
@@ -14,7 +14,6 @@ func GenerateCGAID() (string, error) {
 	partLength := 4
 	totalLength := partLength * 3
 
-	seed := time.Now().UnixNano()
 	rng := rand.New(rand.NewSource(seed)) //nolint: gosec
 
 	// Function to get a random character from allowedChars
@@ -29,7 +28,13 @@ func GenerateCGAID() (string, error) {
 	}
 
 	// Format the custom UID to match CGA(-[23456789cfghjmpqrvwx]{4}){3}
-	formattedUUID := fmt.Sprintf("CGA-%s-%s-%s", randomChars[0:4], randomChars[4:8], randomChars[8:12])
+	formattedID := fmt.Sprintf("CGA-%s-%s-%s", randomChars[0:4], randomChars[4:8], randomChars[8:12])
 
-	return formattedUUID, nil
+	return formattedID, nil
+}
+
+func GenerateCGAID() (string, error) {
+	seed := time.Now().UnixNano()
+
+	return GenerateCGAIDWithSeed(seed)
 }
--- a/pkg/advisory/cga_id.go
+++ b/pkg/advisory/cga_id.go
@@ -1,9 +1,9 @@
 package advisory
 
 import (
-	"crypto/rand"
 	"fmt"
-	"math/big"
+	"math/rand"
+	"time"
 )
 
 func GenerateCGAID() (string, error) {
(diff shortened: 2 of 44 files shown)


pkg/advisory/create.go
```diff
@@ -91,3 +92,38 @@ func createAdvisoryConfig(ctx context.Context, documents *configs.Index[v2.Docum
...
+	// Ensure the string is long enough
+	if len(customUUID) < 12 {
+		return "", fmt.Errorf("hash length is insufficient to generate the UUID with the desired pattern")
+	}
+
+	// Format the custom UUID to match CGA(-[23456789cfghjmpqrvwx]{4}){3}
+	formattedUUID := fmt.Sprintf("CGA-%s-%s-%s", customUUID[0:4], customUUID[4:8], customUUID[8:12])
+
+	return formattedUUID, nil
+}
```
pdeslaur, line 129: Small nit: I'd move this code to a dedicated file so that it's easier to find!

pkg/cli/advisory_create.go
```diff
@@ -147,6 +147,11 @@ newly created advisory and any other advisories for the same package.`,
 				AdvisoryDocs: advisoryCfgs,
 			}
 
+			req.VulnerabilityID, err = advisory.GenerateCGAID(req.Package, req.Aliases[0])
+			if err != nil {
+				return fmt.Errorf("unable to create advisory id: %w", err)
+			}
+
```
luhring, line 154: I think this is a reasonable starting place. We may want to consider handling the ID generation _within_ the lib functions (e.g. `advisory.Create(...)`, called just below), to avoid placing burden on the caller and the risk of them getting this wrong. That would make functions like `advisory.Create` non-deterministic, which maybe is bad, maybe not.

And in either case, I think we'll want to add some check early on in those functions to ensure we got what we're expecting in the `req.VulnerabilityID` field — empty if we make the above change, or filled with a CGA ID if we don't.

luhring (commented):
Nice start! Added some preliminary thoughts

.golangci.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 run:
   concurrency: 6
-  deadline: 5m
+  timeout: 10m
```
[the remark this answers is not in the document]
cpanato: this is more just in case, sometimes ci can delay for any unknown reason

pkg/advisory/create.go
```diff
@@ -91,3 +92,38 @@ func createAdvisoryConfig(ctx context.Context, documents *configs.Index[v2.Docum
 
 	return nil
 }
+
+func GenerateCGAID(packageName, vulnerabilityID string) (string, error) {
```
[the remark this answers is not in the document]
cpanato: agree, one issue we will have will be regarding tests, as we cannot predict the IDs.

pkg/advisory/create.go
cpanato: done

pkg/cli/advisory_create.go
cpanato: move to inside

.golangci.yaml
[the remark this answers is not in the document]
luhring: Okay, gotcha. But I've noticed a strange delay with _golangci-lint in CI_ in particular. Sometimes it even times out (not GHA, but the golangci-lint itself). It'd be great to unearth the root cause before too long...

pkg/advisory/update.go
```diff
@@ -28,13 +26,13 @@ func Update(ctx context.Context, req Request, opts UpdateOptions) error {
 	u := v2.NewAdvisoriesSectionUpdater(func(doc v2.Document) (v2.Advisories, error) {
 		advisories := doc.Advisories
 
-		adv, ok := advisories.Get(vulnID)
+		adv, ok := advisories.GetAlias(req.Aliases[0])
```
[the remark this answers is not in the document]
cpanato: yes! that sounds much better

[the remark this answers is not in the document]
cpanato: will update the code to do that and fallback

pkg/advisory/create.go
```diff
@@ -32,12 +32,15 @@ func Create(ctx context.Context, req Request, opts CreateOptions) error {
...
 	case 1:
-		newAdvisoryID := req.VulnerabilityID
-
 		// i.e. exactly one advisories file for this package
 		u := v2.NewAdvisoriesSectionUpdater(func(doc v2.Document) (v2.Advisories, error) {
-			if _, exists := doc.Advisories.Get(newAdvisoryID); exists {
-				return v2.Advisories{}, fmt.Errorf("advisory %q already exists for %q", newAdvisoryID, req.Package)
+			if _, exists := doc.Advisories.GetAlias(req.Aliases[0]); exists {
+				return v2.Advisories{}, fmt.Errorf("advisory %q already exists for %q", req.Aliases[0], req.Package)
+			}
```
luhring, line 39: I don't think we want to assign special meaning to `req.Aliases[0]` with this new way of doing things. The CGA ID should be the definitive identifier for an advisory; and in contexts like this one ☝️ where the caller doesn't yet know the CGA ID, all aliases (that is, CVE and GHSA IDs for now) should carry equal wait, and we should assume each member of the alias set is equally useful for determining if an advisory already exists to describe this given vulnerability.

pkg/advisory/create.go
```diff
@@ -55,7 +58,7 @@ func Create(ctx context.Context, req Request, opts CreateOptions) error {
 		})
 		err := documents.Update(ctx, u)
 		if err != nil {
-			return fmt.Errorf("unable to create advisory %q for %q: %w", newAdvisoryID, req.Package, err)
+			return fmt.Errorf("unable to create advisory %q for %q: %w", req.Aliases[0], req.Package, err)
```
luhring, line 61: Ditto comment above re: `req.Aliases[0]`, we shouldn't give preference to one alias over others. Here, we could simply show the entire slice.

pkg/advisory/create.go
luhring, line 1: I think we need to update `update.go`, too; is that your understanding?

pkg/advisory/request.go
```diff
@@ -35,8 +36,8 @@ func (req Request) Validate() error {
 		return err
 	}
 
-	if req.VulnerabilityID == "" {
-		return errors.New("vulnerability cannot be empty")
+	if req.VulnerabilityID != "" {
+		return errors.New("vulnerability should be empty")
```
luhring, line 40: This check makes sense to me — but it also almost makes we want to remove the `vulnerabilityID` field from `Request` altogether, if it's only ever a footgun. Curious what you think here!

pkg/configs/advisory/v2/document.go
```diff
@@ -131,13 +131,37 @@ var (
 
 // Get returns the advisory with the given ID. If such an advisory does not
 // exist, the second return value will be false; otherwise it will be true.
-func (advs Advisories) Get(id string) (Advisory, bool) {
+func (advs Advisories) Get(id string, aliases []string) (Advisory, bool) {
```
luhring, line 134: I don't think we need to change this function — we already have `GetByVulnerability`, where the input can be the ID or any alias, right?

pkg/configs/advisory/v2/document.go
```diff
@@ -131,13 +131,37 @@ var (
...
+			}
+		}
+	}
+
+	return Advisory{}, false
+}
+
+// Get returns the advisory with the given ID. If such an advisory does not
+// exist, the second return value will be false; otherwise it will be true.
+func (advs Advisories) GetAlias(id string) (Advisory, bool) {
```
luhring, line 156: Like above, I don't think we should need to add this...

cpanato: yep, i think i forgot to drop this one

pkg/configs/advisory/v2/document.go
cpanato: removed that extra parameter

pkg/advisory/create.go
cpanato: reverted

pkg/advisory/cga_id.go
```diff
@@ -0,0 +1,42 @@
...
+	totalLength := partLength * 3
+
+	// Function to get a random character from allowedChars
+	getRandomChar := func() (byte, error) {
+		max := big.NewInt(int64(len(allowedChars)))
+		n, err := rand.Int(rand.Reader, max)
+		if err != nil {
+			return 0, err
+		}
+		return allowedChars[n.Int64()], nil
```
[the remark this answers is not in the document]
cpanato: updated the code

pkg/advisory/request.go
cpanato: i have no strong opinions here, but let me give a try to remove that, but i think will require more time to refactor the code

the other changes are in place, hope this one is not a blocker for you

pkg/advisory/cga_id.go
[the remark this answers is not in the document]
cpanato: only concern here is that is "less secure"

pkg/advisory/create.go
```diff
@@ -32,12 +32,20 @@ func Create(ctx context.Context, req Request, opts CreateOptions) error {
...
 		// i.e. exactly one advisories file for this package
 		u := v2.NewAdvisoriesSectionUpdater(func(doc v2.Document) (v2.Advisories, error) {
-			if _, exists := doc.Advisories.Get(newAdvisoryID); exists {
-				return v2.Advisories{}, fmt.Errorf("advisory %q already exists for %q", newAdvisoryID, req.Package)
+			if _, exists := doc.Advisories.Get(req.VulnerabilityID); exists {
+				return v2.Advisories{}, fmt.Errorf("advisory %q already exists for %q", req.VulnerabilityID, req.Package)
+			}
+			for _, alias := range req.Aliases {
+				if _, exists := doc.Advisories.Get(alias); exists {
+					return v2.Advisories{}, fmt.Errorf("advisory %q already exists for %q", req.VulnerabilityID, req.Package)
```
luhring, line 42:
```suggestion
					return v2.Advisories{}, fmt.Errorf("advisory %q with alias %q already exists for %q", req.VulnerabilityID, alias, req.Package)
```

luhring: Otherwise I think it would be confusing to the caller as to what went wrong

pkg/advisory/create_test.go
```diff
@@ -164,15 +169,15 @@ func TestCreate(t *testing.T) {
...
+				VulnerabilityID: "CGA-xoxo-xoxo-xoxo",
 			},
 			wantErr: true,
 		},
 		{
 			name: "event type doesn't match data type",
 			req: Request{
 				Package:         "brotli",
-				VulnerabilityID: "CVE-2023-1234",
+				VulnerabilityID: "CGA-xoxo-xoxo-xoxo",
```
luhring, line 180: These will trigger errors (expected, based on `wantErr: true`) but for the wrong reason, right? We don't want the VulnerabilityID to be set. I think it'd be good to remove these values from these test cases, and create a new test case specifically for expecting an error as a result of having a non-empty VulnerabilityID.

pkg/advisory/create_test.go
```diff
@@ -203,8 +208,20 @@ func TestCreate(t *testing.T) {
...
+					// Check if the path is accessing the ID field within the Advisories slice.
+					if len(p) < 2 {
+						return false
+					}
+					if p[len(p)-1].String() == ".ID" {
+						if _, ok := p[len(p)-2].(cmp.SliceIndex); ok {
+							return true
+						}
+					}
+					return false
```
luhring, line 221: This seems a little obscure to me, maybe I just need to look at it for longer. What does that `if _, ok := p[len(p)-2].(cmp.SliceIndex); ok` part do exactly?

luhring (commented):
Hey there — thanks for the changes! 🙇 

This is a partial review. I didn't get time to finish today (e.g. I didn't look at the new alias stuff yet). But I wanted to submit _something_ before I sign off so you're not left with _nothing_ tomorrow morning!

pkg/advisory/cga_id_test.go
```diff
@@ -0,0 +1,31 @@
+package advisory
+
+import (
+	"regexp"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestGenerateCGAID(t *testing.T) {
```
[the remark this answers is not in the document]
cpanato: ok, exposed that function, we can improve that later :)

pkg/configs/advisory/v2/document.go
```diff
@@ -138,6 +138,16 @@ func (advs Advisories) Get(id string) (Advisory, bool) {
...
+	// for _, reqAlias := range aliases {
+	for _, adv := range advs {
+		for _, alias := range adv.Aliases {
+			if alias == id {
+				return adv, true
+			}
+		}
+		// }
+	}
+
```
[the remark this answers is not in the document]
cpanato: i think i was being stupid here 🤦‍♂️

pkg/configs/advisory/v2/document.go
```diff
@@ -169,6 +179,15 @@ func (advs Advisories) Update(id string, advisory Advisory) Advisories {
...
 
+	for i, adv := range advs {
+		for _, alias := range adv.Aliases {
+			if alias == id {
+				advs[i] = advisory
+				return advs
+			}
+		}
+	}
+
```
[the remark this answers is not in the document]
cpanato: see my other comment

pkg/advisory/create_test.go
cpanato: that part was not needed, i got a little help with chatgpt on this, but seems it add things that were not needed 😅

cpanato:
@luhring @pdeslaur did all the feedback review, thanks so much for spending time to review this pr! i think we all are learning a lot with this

pkg/advisory/request.go
luhring: No not a blocker! Good call

pkg/advisory/create.go
luhring: Sorry, just to clarify — `req.Aliases[0]` isn't the entire slice. We can leave this for a later PR if you want just wanted to close the loop!

pkg/advisory/update_test.go
```diff
@@ -135,7 +178,19 @@ func TestUpdate(t *testing.T) {
...
-				if diff := cmp.Diff(tt.expectedDoc, advisoryDocs.Select().WhereName(tt.req.Package).Configurations()[0]); diff != "" {
+				diff := cmp.Diff(tt.expectedDoc, advisoryDocs.Select().WhereName(tt.req.Package).Configurations()[0], cmp.FilterPath(func(p cmp.Path) bool {
+					// Check if the path is accessing the ID field within the Advisories slice.
+					if len(p) < 2 {
+						return false
+					}
+					if p[len(p)-1].String() == ".ID" {
+						if _, ok := p[len(p)-2].(cmp.SliceIndex); ok {
+							return true
+						}
```
luhring, line 189: Did you say we can remove this?

cpanato: removed in the other test and forgot this one, done

pkg/advisory/create.go
cpanato: done to show the entire slice

pkg/configs/advisory/v2/document.go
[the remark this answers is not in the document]
cpanato: i've removed that from the Get method, is that your are talking about?

pdeslaur (approved):

cpanato:
I will update the export osv data in a follow up pr to avoid this become a huge pr

pkg/checks/diff.go
```diff
@@ -299,5 +299,5 @@ func writeDiffLog(diff diffResult, bcz []byte, filename string, newPackages map[
 
 	content := builder.String()
 
-	return os.WriteFile(filename, []byte(content), os.ModePerm)
+	return os.WriteFile(filename, []byte(content), os.ModePerm) //nolint: gosec
```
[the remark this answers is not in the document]
pdeslaur: It looks like it was already like that before 🤔 I'm guessing something changed in the linter. My 2c: I think we could revisit this in a follow-up PR!

pdeslaur (dismissed):

pkg/checks/diff.go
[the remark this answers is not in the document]
cpanato: the rule set `Expect WriteFile permissions to be 0600 or less`
we are setting 777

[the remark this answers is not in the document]
cpanato: did a config for the golangci-lint and update this specific writefile to be 0644

let me know what you think

luhring (approved):

PR #836, opened 2024-05-14
Git based commands, only use GITHUB_TOKEN when interacting with GitHu…

rawlingsj:
…b's API

This change also removes the submodules git update as this is not used anymore, therefore does not need updating with the new GetGitAuth signature.


pkg/git/git.go
```diff
@@ -182,7 +198,10 @@ func TempClone(gitURL, hash string, useAuth bool) (repoDir string, err error) {
 
 	var auth transport.AuthMethod
 	if useAuth {
-		auth = GetGitAuth()
+		auth, err = GetGitAuth(gitURL)
```
tstromberg, line 201: Under what conditions does `TempClone` need authenticated access to GitHub? It seems like read-only usage.

pkg/update/deps/cleanup.go
```diff
@@ -40,13 +40,18 @@ func gitCheckout(p *config.Pipeline, dir string, mutations map[string]string) er
 		return err
 	}
 
+	gitAuth, err := wgit.GetGitAuth(repoValue)
```
tstromberg, line 43: same comment here - do we need to authenticate this request?

tstromberg (changes requested):

pkg/advisory/data_session.go
```diff
@@ -51,10 +51,15 @@ func NewDataSession(ctx context.Context, opts DataSessionOptions) (*DataSession,
...
+	gitAuth, err := wgit.GetGitAuth(opts.Distro.Absolute.AdvisoriesHTTPSCloneURL())
+	if err != nil {
+		return nil, fmt.Errorf("getting git auth: %w", err)
+	}
+
 	// clone advisories repo
 	repo, err := git.PlainCloneContext(ctx, tempDir, false, &git.CloneOptions{
 		URL:  opts.Distro.Absolute.AdvisoriesHTTPSCloneURL(),
-		Auth: wgit.GetGitAuth(),
+		Auth: gitAuth,
```
luhring, line 62: nit: If we save the value of `opts.Distro.Absolute.AdvisoriesHTTPSCloneURL()` to a local variable, we can use that variable down below in the clone options instead of requesting the URL a second time. I like that because it helps us keep the _token_ and the _URL it goes to_ in-sync as we make further changes to the code.

pkg/advisory/data_session.go
```diff
@@ -168,9 +173,14 @@ func (ds DataSession) Modified() bool {
...
-	err := ds.repo.PushContext(ctx, &git.PushOptions{
+	gitAuth, err := wgit.GetGitAuth(ds.distro.Absolute.AdvisoriesHTTPSCloneURL())
+	if err != nil {
+		return fmt.Errorf("getting git auth: %w", err)
+	}
+
+	err = ds.repo.PushContext(ctx, &git.PushOptions{
 		RemoteURL: ds.distro.Absolute.AdvisoriesHTTPSCloneURL(),
-		Auth:      wgit.GetGitAuth(),
+		Auth:      gitAuth,
```
luhring, line 183: (Same comment as above)

pkg/git/git.go
rawlingsj: Looks like the only occurrences of `TempClone`  are called from [advisory diff](https://github.com/wolfi-dev/wolfictl/blob/922b5a2bb4e1915b7f4d05727bdd1f2d1c1a9585/pkg/cli/advisory_diff.go#L42) and [advisory validate](https://github.com/wolfi-dev/wolfictl/blob/922b5a2bb4e1915b7f4d05727bdd1f2d1c1a9585/pkg/cli/advisory_validate.go#L145).  These could be private advisory repos AFAIK which would require auth.  My understanding after looking at the code is that the `Temp` part relates to cloning into a temp dir.  

@luhring can confirm,

pkg/update/deps/cleanup.go
rawlingsj: I think not, @hectorj2f can you confirm and I can remove it in this PR

rawlingsj: I've made this change, will keep the conversation open for when @hectorj2f can agree

pkg/advisory/data_session.go
rawlingsj: 👍

pkg/git/git.go
luhring: That's correct!

pkg/update/deps/cleanup.go
hectorj2f: So we need that git credentials to be able to checkout private repos such as the enterprise-packages.

hectorj2f: I was looking from my phone. I talked with @rawlingsj. This can go away now.

hectorj2f (dismissed):
lgtm

pkg/git/git.go
```diff
@@ -19,20 +22,33 @@ import (
 	gitHttp "github.com/go-git/go-git/v5/plumbing/transport/http"
 )
 
-func GetGitAuth() *gitHttp.BasicAuth {
+func GetGitAuth(gitURL string) (*gitHttp.BasicAuth, error) {
```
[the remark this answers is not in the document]
rawlingsj: sounds good - added

luhring (approved):
LGTM

PR #837, opened 2024-05-14
lint: github.com is not the only git hosting

xnox:
Only require github-update when git checkouts actually are from
github.com rather than from freedesktop gitlab debian salsa etc.

Signed-off-by: Dimitri John Ledkov <dimitri.ledkov@chainguard.dev>

### DIFF
--- a/pkg/lint/rules.go
+++ b/pkg/lint/rules.go
@@ -384,7 +384,7 @@ var AllRules = func(l *Linter) Rules { //nolint:gocyclo
 			Severity:    SeverityError,
 			LintFunc: func(config config.Configuration) error {
 				for _, p := range config.Pipeline {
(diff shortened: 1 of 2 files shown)


pkg/lint/rules.go
```diff
@@ -384,7 +384,7 @@ var AllRules = func(l *Linter) Rules { //nolint:gocyclo
 			Severity:    SeverityError,
 			LintFunc: func(config config.Configuration) error {
 				for _, p := range config.Pipeline {
-					if p.Uses == gitCheckout {
+					if p.Uses == gitCheckout && strings.Contains(p.With["repository"], "github.com") {
```
ajayk, line 387:
```suggestion
					if p.Uses == gitCheckout && strings.HasPrefix(p.With["repository"], "https://github.com/") {
```

ajayk (approved):

Issue #841, opened 2024-05-15
lint: does not know about valid licenses


xnox:
**Description**

2024/05/15 15:44:37 INFO license "GPL-2.0-or-later WITH cryptsetup-OpenSSL-exception" is not valid SPDX license

which is valid, please consider upgrading go-spdx to include https://github.com/github/go-spdx/pull/57

PR #842, opened 2024-05-15
Update go-spdx with more licenses

xnox:
This resolve `wolfictl lint` reporting 4 packages of wolfi-dev/os as
having invalid licenses, when they are, in fact, valid.

Fixes: #841

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -330,3 +330,5 @@ require (
 )
 
 replace golang.org/x/vuln => github.com/luhring/golang-vuln v1.0.2-0.20231029212121-c364fd4725dc
+
+replace github.com/github/go-spdx/v2 v2.2.0 => github.com/darsh12/go-spdx/v2 v2.0.0-20240506204000-9af17522cc6e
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -330,3 +330,5 @@ require (
 )
 
 replace golang.org/x/vuln => github.com/luhring/golang-vuln v1.0.2-0.20231029212121-c364fd4725dc
+
+replace github.com/github/go-spdx/v2 v2.2.0 => github.com/darsh12/go-spdx/v2 v2.0.0-20240506204000-9af17522cc6e
```
imjasonh, line 334: Is this fork proposed anywhere as a PR on the main repo? If not, is there an issue describing why it wouldn't be?

I'm a little nervous about relying on an unknown maintainer for this. Not because of vulns (it's pinned and looks sane), but more for maintenance burden going forward. If go-spdx adds features, will this fork integrate them?

xnox: @imjasonh it is proposed here https://github.com/github/go-spdx/pull/57 i don't know how else we can pull it.

But basically there are a couple of licenses that we use; and are approved; but unknown. The alternative is for me to add them to exceptions like I did for custom in the other pr.

It is like cryptsetup WITH cryptsetup-OpenSSL exception and Boehm-GC.

imjasonh: Thanks that's helpful!

imjasonh (approved):

PR #857, opened 2024-05-16
build: use the other gcp auth plugin

imjasonh:
The GCP auth plugin moved out-of-tree in client-go, so let's use the new location.


xnox:
This will likely conflict with https://github.com/wolfi-dev/wolfictl/pull/856

xnox (approved):

k4leung4 (approved):

PR #865, opened 2024-05-21
update osv export to check if there is no id colision and drop the id placeholder

cpanato:
- check if there is no id collision when exporting 
- drop the id placeholder as prefix for the `ID` and filename
- update `all.json` to have only the `id` and `modified` fields 

xref: https://github.com/wolfi-dev/wolfictl/pull/835


cpanato:
need #835 then i can update the tests

rawlingsj (approved):

hectorj2f (approved):

PR #883, opened 2024-05-23
advisory copy: method to bootstrap an advisory YAML from a previous version stream advisory

imjasonh:


pkg/cli/advisory_copy.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/spf13/cobra"
+	v2 "github.com/wolfi-dev/wolfictl/pkg/configs/advisory/v2"
+	rwos "github.com/wolfi-dev/wolfictl/pkg/configs/rwfs/os"
+)
+
+func cmdAdvisoryCopy() *cobra.Command {
+	var dir string
+	cmd := &cobra.Command{
+		Use:   "copy",
```
luhring, line 16: Since these aren't flags with names, maybe we update the usage to spell out what the positional args do? 🤷 

```suggestion
		Use:   "copy <source-package-name> <destination-package-name>",
```

luhring (approved):

PR #900, opened 2024-05-27
bundles: Split APK upload from indexing

jonjohnsonjr:
Since we're taking out a lock to update the index, we should move the APK uploading outside of that locked section.

### DIFF
--- a/pkg/cli/build.go
+++ b/pkg/cli/build.go
@@ -1107,10 +1107,45 @@ func (t *task) buildBundle(ctx context.Context) error {
 		return nil
 	}
 
+	log.Infof("Processing results: %s", t.pkg)
+
+	tmpdir, err := os.MkdirTemp("", "")
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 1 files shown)


pkg/cli/build.go
```diff
@@ -1107,11 +1107,46 @@ func (t *task) buildBundle(ctx context.Context) error {
...
+	tmpdir, err := os.MkdirTemp("", "")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(tmpdir)
+
+	var filemu sync.Mutex
+	filesByArch := map[string][]string{}
+
+	var pkgGroup errgroup.Group
```
imjasonh, line 1121: WithContext?

imjasonh (approved):

pkg/cli/build.go
jonjohnsonjr: I think not -- even if one arch fails, we still want to upload the other arch's packages.

PR #927, opened 2024-06-03
apk ls: list repository packages

imjasonh:
This incorporates `apkrane ls` to list packages in a repo, optionally in JSON and optionally only latest packages, etc.

This supports HTTP_AUTH, local and stdin indexes.

### DIFF
--- a/pkg/cli/apk.go
+++ b/pkg/cli/apk.go
@@ -28,7 +28,7 @@ func cmdApk() *cobra.Command {
 	return cmd
 }
 
-func cmdCp() *cobra.Command { //nolint:gocyclo
+func cmdCp() *cobra.Command {
 	var latest bool
 	var indexURL, outDir, gcsPath string
 	cmd := &cobra.Command{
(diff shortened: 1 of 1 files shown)


jonjohnsonjr (approved):

pkg/cli/apk.go
```diff
@@ -221,14 +189,124 @@ func cmdCp() *cobra.Command { //nolint:gocyclo
...
+
+func fetchAPKIndex(ctx context.Context, indexURL string) (*apk.APKIndex, string, error) {
+	var arch string
+	repoURL := strings.TrimSuffix(indexURL, "/APKINDEX.tar.gz")
+	var in io.ReadCloser
+	switch {
+	case indexURL == "-":
+		in = os.Stdin
+		arch = "x86_64" // TODO: This is hardcoded.
+	case strings.HasPrefix(indexURL, "file://"):
```
luhring, line 264: I know these were here before, just curious if we use these, or just have them here because the seem good?

luhring (approved):

pkg/cli/apk.go
imjasonh: Mostly just because they seem good. I'm not married to it if we decide to drop it now/later.

PR #943, opened 2024-06-06
bundles: Fix race in obj generation updates

jonjohnsonjr:
We made index updates a lot faster, which exposed a race conditions in how we track APKINDEX object generations with concurrent map updates.


pkg/cli/build.go
```diff
@@ -243,7 +243,9 @@ func (g *global) fetchIndexFromBucket(ctx context.Context, arch string) (map[str
 	defer rc.Close()
 
 	// Set this for conditional requests on upload.
+	g.genmu.Lock()
```
k4leung4, line 246: is a sync.Map be better than explicit locks?

imjasonh (approved):

PR #948, opened 2024-06-10
tests: add SBOM integration tests (for Syft)

luhring:
This gives us a way to catch problems coming from Syft....... and it caught some!

We'll investigate these async to this PR, but there is nondeterministic behavior causing the outcome of the tests for `jenkins` and `jruby` to be unstable. 😱


imjasonh (approved):
🎉❤️🙌

pkg/sbom/sbom_test.go
```diff
@@ -0,0 +1,155 @@
...
+func TestGenerate(t *testing.T) {
+	const wolfiAPKRepoURL = "https://packages.wolfi.dev/os"
+
+	testArtifacts := []string{
+		"crane-0.19.1-r6.apk",
+
+		// Note: Syft is producing nondeterministic results for these APKs, so the tests
+		// aren't stable. Uncomment them when the issue is resolved.
+		//
+		// "jenkins-2.461-r0.apk",
```
vaikas, line 34: Is there an issue tracking this that we could include here so we know how to keep track of this work?

luhring: Great callout — I'll open an issue and add the reference to the comment here when I open this PR's companion for vuln scan integration tests.

PR #958, opened 2024-06-15
Add integration tests for vulnerability scanning

luhring:
🚀  This adds integration tests to catch any unexpected behavior in the vulnerability scanning code in wolfictl.

⚠️  This also introduces a **breaking change** in the scanner library API, which was necessary to get the scanning logic to be more testable. The change was to create a new `ScannerOptions` struct as the sole parameter in the scanner constructor, so the hope is that we don't need to make breaking changes as we further extend the scanner constructor options going forward. cc: @cpanato @hectorj2f 

⬆️  This bumps Syft and Grype to their latest versions.


luhring:
Needs https://github.com/anchore/grype/pull/1950

luhring:
@hectorj2f Note that this also bumps Grype to >= `v0.79.0`, which turns off the Go main module 0.0.0 matching by default (I had told you I'd ping you when this happened!)

hectorj2f (commented):
lgtm

imjasonh (approved):
LGTM, thanks for the incredible docs for this ❤️

PR #966, opened 2024-06-18
refactor: separate OSV build from export command

luhring:
This is a relatively straightforward separation of the OSV dataset build logic from the existing `export` command.

The `export` command is marked **experimental** and not designed to be load-bearing.

Meanwhile, we intend for our OSV output to be a **first-class, production-ized output**, on par with our `secdb` generation, which also has its own command and implementation (distinct from `export`).

This PR establishes a new entrypoint for OSV dataset builds: `wolfictl adv osv`. The core implementation logic has not been modified yet.

cc: @imjasonh @cpanato

### DIFF
--- a/pkg/advisory/export.go
+++ b/pkg/advisory/export.go
@@ -3,19 +3,11 @@ package advisory
 import (
 	"bytes"
 	"encoding/csv"
-	"encoding/json"
 	"fmt"
 	"io"
-	"log"
-	"os"
-	"path"
-	"sort"
-	"strings"
-	"time"
 
 	_ "github.com/santhosh-tekuri/jsonschema/v5/httploader" // to be able to download the schema from the URL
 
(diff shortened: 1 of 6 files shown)


pkg/advisory/export.go
```diff
@@ -126,153 +115,3 @@ func ExportYAML(opts ExportOptions) (io.Reader, error) {
...
-		documents := index.Select().Configurations()
-
-		for _, doc := range documents {
-			for _, adv := range doc.Advisories {
-				sortedEvents := adv.SortedEvents()
-
-				var updatedTime time.Time
-				tempAffected := models.Affected{}
-
-				for _, event := range sortedEvents {
```
jdolitsky, line 143: NOW THATS A LOOP!

jdolitsky (approved):

PR #967, opened 2024-06-19
Upgrade melange to fix git checkout with submodules

xnox:
Signed-off-by: Dimitri John Ledkov <dimitri.ledkov@chainguard.dev>


jonjohnsonjr:
Fixed by https://github.com/wolfi-dev/wolfictl/pull/970

PR #988, opened 2024-06-21
feat(osv): account for subpackages and the Wolfi ecosystem

luhring:
This PR continues to improve the OSV tests and makes the following important changes:

1. Since we're moving toward a unified OSV feed, we didn't have a way to account for the duality of a given advisory's affect on a package that scanners might think of both as a "Wolfi" package and a "Chainguard" package. This is because a package that comes from Wolfi may be included in an image that has either the distro ID of `wolfi` or `chainguard`. Now, for advisories affecting such a package, we account for both ecosystems, to ensure scanners find a match to this advisory when they should.
2. We hadn't been listing subpackages as affected by any advisory. With the secdb (our main feed prior to introducing OSV data), we document that the scanner is responsible for mapping a subpackage to its origin package (using APK metadata available in the scanned artifact) in order to lookup applicable vulnerabilities. With OSV, we're making this more explicit for scanners so no implicit mapping is needed. Scanners can take any identified subpackages and use those directly when looking up vulnerabilities using Chainguard's provided OSV data.

⚠️ This necessitates breaking changes to the CLI. The updated CLI expectations are documented in `wolfictl adv osv --help`. @cpanato Let's coordinate the rollout of this update at the start of next week to make sure we're both happy with it!


pkg/advisory/osv.go
```diff
@@ -21,119 +21,143 @@ import (
...
+	for _, packageConfigIndex := range opts.PackageConfigIndices {
+		// We assume that each package name is unique across all package repos.
+
+		cfgs := packageConfigIndex.Select().Configurations()
+		for i := range cfgs {
+			cfg := cfgs[i]
+
+			var subpackages []string
+			for i := range cfg.Subpackages {
+				sp := cfg.Subpackages[i]
```
jdolitsky, line 76:
```suggestion
			for _, sp := range cfg.Subpackages {
```
?

pkg/cli/advisory_osv.go
```diff
@@ -3,42 +3,92 @@ package cli
...
+				packageIndices = append(packageIndices, index)
+			}
+
+			addedEcosystems := []string{"Wolfi"}
+			for i := range p.advisoriesRepoDirs {
+				if i == 0 {
+					// Skip the first advisory repository, which is the "Wolfi" repository.
+					continue
+				}
+				addedEcosystems = append(addedEcosystems, "")
```
jdolitsky, line 84: what does empty string mean here?

pkg/advisory/osv.go
```diff
@@ -21,119 +21,143 @@ import (
...
+	// Do one time upfront, instead of per advisory document: Find out the
+	// subpackages for all defined packages in the packages repos.
+	pkgNameToSubpackages := make(map[string][]string)
+	for _, packageConfigIndex := range opts.PackageConfigIndices {
+		// We assume that each package name is unique across all package repos.
+
+		cfgs := packageConfigIndex.Select().Configurations()
+		for i := range cfgs {
+			cfg := cfgs[i]
+
```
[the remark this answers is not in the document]
luhring: I actually had it that way originally, but the linter said it was copying too many bytes each time and to consider doing it that way ☝️

pkg/advisory/osv.go
luhring: See above

pkg/cli/advisory_osv.go
luhring: I kind of explain it here but I could do a better job — https://github.com/luhring/wolfictl/blob/5372e24fc044fc9787d0566809d415ff864d8605/pkg/advisory/osv.go#L35-L43

I'll update that. I'm hoping this is short-lived, we just need a way to sometimes add an extra ecosystem depending on which advisory dataset we're using.

jdolitsky (approved):

pkg/cli/advisory_osv.go
luhring: a2b5229

PR #989, opened 2024-06-21
DNM: feat(adv/validate): check for package config existing

luhring:
This introduces a new advisory validation check to make sure each advisory document describes a package that's currently defined in the corresponding package repo (e.g. in `wolfi-dev/os`, if the advisories repo is `wolfi-dev/advisories`).

The goal is to help us move toward a state where each defined package has only one advisory document across the universe of advisory data. This will make it easier for us to unify all advisory data into a single repo. And before that point, it also helps ensure we don't have duplicate or stale data that can leak into our various downstream data artifacts (e.g. the secdb, the OSV feed, and other export destinations).

Left to do:
- [ ] Tests
- [ ] Docs


luhring:
closing because old

PR #997, opened 2024-06-24
apk ls --newer-than: show only packages built recently

imjasonh:
\# of latest packages built in the last 2 weeks:
```
$ go run ./ apk ls https://packages.wolfi.dev/os/x86_64/APKINDEX.tar.gz --latest --newer-than 336h | wc -l
    1783
```

vs last 24 hours:
```
$ go run ./ apk ls https://packages.wolfi.dev/os/x86_64/APKINDEX.tar.gz --latest --newer-than 24h | wc -l
      49
```

This can be done with some `jq`-fu, but that's tedious and error prone.

### DIFF
--- a/pkg/cli/apk.go
+++ b/pkg/cli/apk.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"slices"
 	"strings"
+	"time"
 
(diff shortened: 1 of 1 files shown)


smoser (approved):
I like it.  Thank you.

PR #1034, opened 2024-07-06
validate: advisory: file and package name match

Dentrax:
This PR adds a new validation step for advisory. Package name should also reflect the actual file name to work as expected. For example; package name should reflect the version stream in the advisories. `foo-1.2.advisories.yaml`'s `package.name` MUST be `foo-1.2` rather than `foo`.


luhring (changes requested):
🎉  Wow, this is a great enhancement, and we really should've had it already! Thank you!!

There's a simpler and smaller-scoped way to achieve the same thing.

The `configs.Index[v2.Document]` type actually knows the file paths that correspond to every advisory document already. It'd be nice not to have to change the shape of the `v2.Document` struct itself just to be able to implement this validation check, as that's requiring you to make even more changes in this PR to satisfy tests and what not.

The way the "configs index" thing works is that it has a bunch of `Entry`s, and each `Entry` holds one `v2.Document` (or whatever the instantiated generic type is at the moment). So while the `v2.Document` doesn't (and shouldn't, IMHO) know about its backing file path, the `Entry` that wraps it does!

In fact, there's already a [method](https://github.com/wolfi-dev/wolfictl/blob/998fbe4010831f4973940124b0a48b409d528ac3/pkg/configs/entry.go#L22) on `Entry` to get the file path, but it's **private** 😢 . It should really be public, I think, to help enable both this PR and other future features. **I can make this change in a quick PR**, and then you should be able to rebase, and the only change to the `advisory.Validate` function you should need to make is to add something like this:

```go
var documentNameErrs []error
for _, e := range opts.AdvisoryDocs.Select().Entries() {
	if e.Path() != e.Configuration().Package.Name+".advisories.yaml" {
		documentNameErrs = append(documentNameErrs, fmt.Errorf("document file name %q does not match package name %q", e.Path(), e.Configuration().Package.Name))
	}
}
errs = append(errs, errorhelpers.LabelError("document file name validation failure(s)", errors.Join(documentNameErrs...)))
```

Dentrax:
Thank you, @luhring, for your PR and enlightening my way here! The new solution is simpler and nicer now. PTAL when possible! 💜

luhring (approved):
Nice work! 🎉

PR #1042, opened 2024-07-08
advisory: stream command

Dentrax:
Start version streaming for a package by moving its advisories into a new package.

This command will move most advisories for the given package into a new package. And rename the
package to the new package name. (i.e., from foo.advisories.yaml to foo-X.Y.advisories.yaml) If the
target file already exists, the command will try to merge the advisories. To ensure the advisories
are up-to-date, the command will start a scan for the new package.

The command will move the latest event for each advisory, and will update the timestamp
of the event to now. The command will not copy events of type "detection", "fixed",
"analysis_not_planned", or "fix_not_planned".

---

Test it:
```
1. `go install .`
2. Jump to https://github.com/wolfi-dev/advisories/ dir
3. Run `wolfictl advisory stream kyverno kyverno-1.12`
```

### DIFF
--- a/pkg/cli/advisory.go
+++ b/pkg/cli/advisory.go
@@ -48,6 +48,7 @@ func cmdAdvisory() *cobra.Command {
 		cmdAdvisorySecDB(),
 		cmdAdvisoryUpdate(),
 		cmdAdvisoryValidate(),
+		cmdAdvisoryStream(),
 	)
 
 	return cmd
--- a/pkg/cli/advisory_copy.go
+++ b/pkg/cli/advisory_copy.go
@@ -49,39 +49,9 @@ of the event to now. The command will not copy events of type "detection", "fixe
 			out.Advisories = nil
 
 			for _, adv := range hdoc.Advisories {
-				evts := make([]v2.Event, 0, len(adv.Events))
-
-				for _, evt := range adv.Events {
-					switch evt.Type {
-					case v2.EventTypeDetection, v2.EventTypeFixed, v2.EventTypeAnalysisNotPlanned, v2.EventTypeFixNotPlanned:
-						// Don't carry these over.
-						continue
-
-					case v2.EventTypePendingUpstreamFix, v2.EventTypeFalsePositiveDetermination, v2.EventTypeTruePositiveDetermination:
-						// Carry these over as-is.
-						evts = append(evts, evt)
-
-					default:
-						// A new type was added and we don't know how to handle it. Default to not carrying it over.
-					}
+				if carried, ok := carryAdvisory(adv); ok {
+					out.Advisories = append(out.Advisories, carried)
 				}
-
-				if len(evts) == 0 {
-					// No events to carry over.
-					continue
-				}
-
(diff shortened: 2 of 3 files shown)


luhring:
Thanks for doing this, @Dentrax! This looks like it will be useful.

One overall question before diving in too deep: would this be more useful as a generalized "move"/"rename" command? It sounds like one use case is when we start to version stream a package that wasn't version streamed before, like the kyverno example. But IIUC, the same logic applies for any package rename — does that sound right to you?

I'm wondering if by generalizing this command to something like `wolfictl advisory mv` (inspired by and analogous to the Unix `mv` command, which handles file renames too), and keeping almost all of the logic in this PR (minus the version stream regex checking), we'd solve not just the "new version stream" case but all package rename cases... and also be well on our way to a "moved across repos" solution as well. What are your thoughts? Feel free to tell me I'm misunderstanding things here. 😄

Dentrax:
Great advice @luhring! Renamed to `move (mv)`. 👍

luhring:
Would you mind looking at the code a bit again before we review further? It looks like there is still some version stream logic left over, documentation files from the old command name, and references to scanning that we ended up deciding not to do. Once this is in a state that's ready for more scrutiny, just ping me!

Dentrax:
Apologies for the oversight @luhring! I should have double-check it.

It's now ready for review. PTAL when possible!

> and references to scanning that we ended up deciding not to do.

I couldn't find this, actually i didn't include this functionality. 🤔

docs/cmd/wolfictl_advisory_move.md
```diff
@@ -0,0 +1,47 @@
...
+```
+
+### Synopsis
+
+Move a package's advisories into a new package.
+
+This command will move most advisories for the given package into a new package. And rename the
+package to the new package name. (i.e., from foo.advisories.yaml to foo-X.Y.advisories.yaml) If the
+target file already exists, the command will try to merge the advisories. To ensure the advisories
+are up-to-date, the command will start a scan for the new package.
```
luhring, line 20: It looks like this PR still needs some cleanup from our last discussion?

pkg/configs/index_test.go
```diff
@@ -32,3 +32,39 @@ func TestNewIndex(t *testing.T) {
...
+	require.NoError(t, err)
+
+	_, err = index.Select().WhereName(name).First()
+	require.NoError(t, err)
+
+	t.Run("removes a config", func(t *testing.T) {
+		err := index.Remove(filename)
+		require.NoError(t, err)
+
+		assert.NotContains(t, index.paths, filename)
```
luhring, line 63: What is this line for? It looks like we're testing the removal in the test block below?

pkg/configs/index_test.go
```diff
@@ -32,3 +32,39 @@ func TestNewIndex(t *testing.T) {
...
+	})
+
+	t.Run("ensure the config is removed", func(t *testing.T) {
```
luhring, line 66: Should we also test that other configs _weren't_ removed and are still available?

PR #1044, opened 2024-07-09
set build ID, update melange, use apko auth

imjasonh:
- update `chainguard.dev/melange@main` to pick up vars fixes
- generate a UUID each time `wolfictl build` is run, and pass it through as a pod annotation
- replace `addAuth` with apko's `auth.DefaultAuthenticators.AddAuth`.
- support `-a` to pass arbitrary `key=value` annotations to the pod spec (e.g., GH run ID, etc.)


k4leung4 (dismissed):

pkg/private/bundle/bundle.go
```diff
@@ -497,9 +484,11 @@ func Podspec(task Task, ref name.Reference, arch, mFamily, sa, ns string, gvisor
...
 	}
 
-	if gvisor {
-		c := "gvisor"
-		pod.Spec.RuntimeClassName = &c
+	for k, v := range anns {
+		if _, ok := pod.Annotations[k]; ok {
+			return nil, fmt.Errorf("annotation %q already set", k)
+		}
+		pod.Annotations["melange.chainguard.dev/"+k] = v
```
jonjohnsonjr, line 491: Above checks `k` but this sets `"melange.chainguard.dev/"+k` is that intentional?

imjasonh: Not intentional, fixed.

jonjohnsonjr (approved):

PR #1047, opened 2024-07-09
fix(osv): move aliases to related field

luhring:
See https://github.com/google/osv.dev/issues/2374 for motivation.

cc: @cpanato

### DIFF
--- a/pkg/advisory/osv.go
+++ b/pkg/advisory/osv.go
@@ -148,8 +148,9 @@ func BuildOSVDataset(_ context.Context, opts OSVOptions) error {
 					continue
 				}
 
-				// Note: The OSV data should include our advisory ID itself among the listed aliases.
-				aliases := append([]string{adv.ID}, adv.Aliases...)
+				// Note: The OSV data should include our advisory ID itself among the listed
+				// related vulnerability IDs.
+				related := append([]string{adv.ID}, adv.Aliases...)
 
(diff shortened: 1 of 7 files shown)


imjasonh (approved):

pkg/advisory/osv.go
```diff
@@ -148,8 +148,9 @@ func BuildOSVDataset(_ context.Context, opts OSVOptions) error {
 					continue
 				}
 
-				// Note: The OSV data should include our advisory ID itself among the listed aliases.
-				aliases := append([]string{adv.ID}, adv.Aliases...)
+				// Note: The OSV data should include our advisory ID itself among the listed
+				// related vulnerability IDs.
+				related := append([]string{adv.ID}, adv.Aliases...)
```
michaelkedar, line 153: Hi,
I just wanted to ask why you include the advisory's own ID in its `related` (or `aliases`) field.
It seems a bit odd to make a record link to itself, but it seem like an intentional decision?

cpanato: was my understanding in the beginning to add that as is. maybe i got that wrong

luhring: Yeah it sounded like this was feedback from the OSV team — originally we didn't add in the ID itself to this collection. Happy to remove it if it's not needed! 🧹

PR #1066, opened 2024-07-15
lint: Only fail command if error is returned.

wlynch:
If errors only contain warnings, treat this as a success.


wlynch:
cc @rawlingsj

pnasrat (approved):
LGTM

PR #1077, opened 2024-07-24
Option in adv guide to print yaml advisory to console

jdolitsky:
It seems that the "submit PR" function requires a GITHUB_TOKEN (correct?) so this PR allows the user to simply print the changed YAMLs to the console to submit to advisory repo themselves.

### DIFF
--- a/pkg/advisory/request.go
+++ b/pkg/advisory/request.go
@@ -36,9 +36,9 @@ func (req Request) Validate() error {
 		return err
 	}
 
-	if req.VulnerabilityID != "" {
-		return errors.New("vulnerability should be empty")
-	}
+	// if req.VulnerabilityID != "" {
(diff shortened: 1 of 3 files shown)


pkg/advisory/request.go
```diff
@@ -36,9 +36,9 @@ func (req Request) Validate() error {
 		return err
 	}
 
-	if req.VulnerabilityID != "" {
-		return errors.New("vulnerability should be empty")
-	}
+	// if req.VulnerabilityID != "" {
+	// 	 return errors.New("vulnerability should be empty")
+	// }
```
jdolitsky, line 42: This seems to cause `wolfictl adv guide` to crash with
```
2024/07/23 21:35:03 ERRO adding advisory data: creating advisory: vulnerability should be empty
```

PR #1119, opened 2024-08-08
fix: add support to clone using the branch

hectorj2f:
related to this error: https://github.com/wolfi-dev/os/actions/runs/10289224641/job/28476459222?pr=25816

and this discussion on Slack: https://chainguard-dev.slack.com/archives/C0627BW5GLF/p1723132481217119

### DIFF
--- a/pkg/update/deps/cleanup.go
+++ b/pkg/update/deps/cleanup.go
@@ -30,7 +30,7 @@ func gitCheckout(p *config.Pipeline, dir string, mutations map[string]string) er
 	}
 
 	tagValue := p.With["tag"]
-	if tagValue == "" {
+	if tagValue == "" && p.With["branch"] == "" {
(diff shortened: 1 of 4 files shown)


imjasonh (approved):

pkg/update/deps/cleanup.go
```diff
@@ -30,7 +30,7 @@ func gitCheckout(p *config.Pipeline, dir string, mutations map[string]string) er
 	}
 
 	tagValue := p.With["tag"]
-	if tagValue == "" {
+	if tagValue == "" && p.With["branch"] == "" {
```
luhring, line 33: nit: I might extract the expression `p.With["branch"]` into a value since we end up accessing it 3 times here

luhring (approved):

PR #1126, opened 2024-08-09
build(deps): bump syft,grype; update golden files

luhring:


pkg/sbom/testdata/goldenfiles/aarch64/jenkins-2.461-r0.apk.syft.json
```diff
@@ -83520,15 +83548,15 @@
...
           "artifactId": "http2-common",
-          "version": "",
+          "version": "10.0.20",
           "name": "Jetty :: HTTP2 :: Common"
         }
       }
     },
     {
-      "id": "b63e201ee1e49a0d",
+      "id": "02e0b2afd4c6d0e3",
```
imjasonh, line 83559: Are these IDs going to change every time we bump Syft? What are they?

If they're toilsome we can make the diff ignore them instead.

imjasonh (approved):

pkg/sbom/testdata/goldenfiles/aarch64/jenkins-2.461-r0.apk.syft.json
luhring: They're a deterministic hash of most of the other fields. They'll change if the package changes meaningfully, so I personally don't mind them or think we should factor them out of the diff.

PR #1104, opened 2024-08-01
remove bundle code

imjasonh:
This change reduces `wolfictl`'s size from 120MB -> 84MB

PR #1127, opened 2024-08-11
don't depend on apko's custom log package

imjasonh:
Depends on https://github.com/chainguard-dev/melange/pull/1430


k4leung4 (dismissed):

pkg/cli/build.go
```diff
@@ -112,7 +112,6 @@ func cmdBuild() *cobra.Command {
 	cmd.Flags().StringVar(&cfg.cacheSource, "cache-source", "", "directory or bucket used for preloading the cache")
 	cmd.Flags().BoolVar(&cfg.generateIndex, "generate-index", true, "whether to generate APKINDEX.tar.gz")
 	cmd.Flags().StringVar(&cfg.DestinationRepo, "destination-repository", "", "repo used to check for (and skip) existing packages")
-	cmd.Flags().StringVar(&cfg.UploadRepo, "upload-repository", "", "repo where packages will be uploaded")
```
luhring, line 115: Just double checking this since I didn't see it in the description — we just don't need this any more?

pkg/cli/commands.go
```diff
@@ -23,7 +22,6 @@ func New() *cobra.Command {
 			return nil
 		},
 	}
-	cmd.PersistentFlags().StringSliceVar(&logPolicy, "log-policy", []string{"builtin:stderr"}, "log policy (e.g. builtin:stderr, /tmp/log/foo)")
```
luhring, line 26: Nice!

luhring (approved):

pkg/cli/build.go
imjasonh: Yep, that was only used by the bundle stuff, which was removed in #1104

PR #1150, opened 2024-08-27
Consider subpackage runtime deps for graph

jonjohnsonjr:
We were only looking at main package runtime deps when constructing the dependency graph, which meant we missed a lot of edges that were very important.

### DIFF
--- a/pkg/cli/dot.go
+++ b/pkg/cli/dot.go
@@ -64,6 +64,11 @@ Open browser to explore crane's deps recursively, only showing a minimum subgrap
 				return fmt.Errorf("building graph: %w", err)
 			}
 
+			g, err = g.Targets()
+			if err != nil {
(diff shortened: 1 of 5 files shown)


jonjohnsonjr:
Blah. When I run this locally:

```
wolfictl text --type=name \
    --pipeline-dir=./pipelines/
```

No errors,  but when I run this:
```
wolfictl text --type=name \
    --pipeline-dir=./pipelines/ \
    --keyring-append=https://packages.wolfi.dev/bootstrap/stage3/wolfi-signing.rsa.pub \
    --repository-append=https://packages.wolfi.dev/bootstrap/stage3
```

Many errors. That is a little surprising to me.

smoser:
https://github.com/wolfi-dev/os/pull/27275 should fix the py3-setuptools-scm

imjasonh (approved):

PR #1153, opened 2024-08-31
Fix for invalid version in recreated package files bug

jamie-albert:
draft until validation.go tests can be implemented


luhring:
Left some thoughts in Slack — let me know if you want to jump on a call or something to chat! 😃

PR #1154, opened 2024-09-03
feat(adv): add basic logging to osv data generation

luhring:
Also fix `newLogger` to use charm logging better.

Example of new logging:

image


luhring:
This is intended to make it easier to spot problems in our OSV generation workflow.

jdolitsky (approved):
bufo bless

cpanato (commented):
late lgtm

thanks

PR #1159, opened 2024-09-03
Removing fixed-version-present-and-first logic and test

jamie-albert:
As per conversation with @luhring, We agree that the stringent rule of the fixed version of a package can not be the first version should be relaxed and addressed more methodically moving forward. An example of where this is in the instance where validation is recommending to change previously valid fix version tags [to 'false-positive-determination'](https://github.com/chainguard-dev/enterprise-advisories/actions/runs/10622587699/job/29447212893#step:7:27) due to the commit history being lost when the package file was deleted and remade.

### DIFF
--- a/pkg/advisory/validate.go
+++ b/pkg/advisory/validate.go
@@ -5,11 +5,8 @@ import (
 	"errors"
 	"fmt"
 	"slices"
-	"sort"
 	"strings"
 	"time"
-
-	goapkversion "github.com/knqyf263/go-apk-version"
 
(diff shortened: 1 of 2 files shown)


luhring (approved):
LGTM!

My hope is soon we can upgrade the schema to describe exactly how any given fix is claimed to have been made, and an auditable history of APK existence in the APK registry, both of which should help consumers of the data determine for themselves whether a given APK version fixed a pre-existing vulnerability or not (and how).

PR #1166, opened 2024-09-05
fix(adv/cp): new advisories need unique CGA IDs

### LINKED ISSUE #1165: `wolfictl adv cp` reuses the same CGA ID, which is invalid
CGA IDs should be unique to the vulnerability+package tuple, so when we're copying existing advisory data from one package to another, we need to take care to generate a new CGA ID in the process. Today, the command produces data that fails advisory validation.

This was surfaced by @smoser!

luhring:
Fixes #1165


smoser:
Thanks :+1:

PR #699, opened 2024-03-14
secdb: ignore when you find multiple advisories across os and other r…

ajayk:
…epos for now

PR #1185, opened 2024-09-15
fix(adv): validate fixed versions only for new data

luhring:
Previously, the validation logic made an assumption that removing packages from the APKINDEX is an exceptional and rare event, used only when a severe mistake is made during a package build. Since the logic's introduction, this assumption has become invalid, as there are more common cases where packages are removed from the distro now.

Consequently, this PR moves validation of "fixed" version existence to focus on only new fixed versions being added to the advisory data, as opposed to globally looking at all advisory data and the fixed versions mentioned therein.

⚠️ This also partially reverts the work done in https://github.com/wolfi-dev/wolfictl/pull/699, which I believe was incorrect and leading indirectly to a unit test failure. I'll add a review comment to the code diff to make sure I'm making the right change here.


pkg/advisory/validate.go
```diff
@@ -266,11 +182,6 @@ func (opts ValidateOptions) validateBuildConfigurationExistence(pkgName string)
 	}
 
 	if _, ok := opts.distroPackageMap[pkgName]; !ok {
-		// Double check if the package is in index and move on
-		if _, ok := opts.apkIndexPackageMap[pkgName]; ok {
-			// the package build configuration has moved out but not removed from the index or apks
-			return nil
-		}
```
luhring, line 273: @ajayk Hi! I just wanted to get you to sanity check me here, since this is undoing a few lines of your work. The logic removed here was allowing the advisory data to refer to a package that's not defined as a Melange file in the distro, as long as it's listed in the APKINDEX.

But `validateBuildConfigurationExistence` is only called for **net new documents** — so the data that's being added _right now_ by the user.

You had a explanation comment here that I'm also removing:

```go
// This would change from March 13,2024 as package yamls would be deleted for EOL and non latest versions
// Only the package (apk) versions would exists in wolfi and wolfi index but no build config
```

If we're _deleting_ documents, I don't believe this logic actually comes into play at all. And I think we do want to make sure that documents being added in a PR **_are_** referencing a package that's still in the distro.

Curious for your thoughts here!

pkg/advisory/validate.go
```diff
@@ -517,6 +416,25 @@ func (opts ValidateOptions) validateIndexDiff(ctx context.Context, diff IndexDif
 	return errorhelpers.LabelError("invalid change(s) in diff", errors.Join(errs...))
 }
 
+func (opts ValidateOptions) validateIndexDiffForAddedEvents(events []v2.Event, packageName string) error {
```
hectorj2f, line 419: Do we have tests to validate this change ?

luhring: Good callout... some of the behavior is implicitly asserted already, but it makes sense to add more explicit conditions — added in 7452370

hectorj2f (approved):

hectorj2f:
@luhring Could we merge this PR ?

luhring:
> @luhring Could we merge this PR ?

Good with me — just also see above review comment for my sanity check question meant for @ajayk.

luhring:
I'm going to merge this now! If anyone notices an issue from these changes, just holler!

PR #1251, opened 2024-10-18
disable FromContent for the go cataloger

imjasonh:


imjasonh:
Sure enough, this causes our SBOMs to produce bad main module versions, as evidenced by the change to the thanos goldenfile required to make its test pass.

luhring:
(yay for tests!) thanks for trying this out nonetheless!

PR #1281, opened 2024-10-29
text: Only consider local repo for self-providers

jonjohnsonjr:
This was seemingly added to prevent accidentally having cycles in the dependency graph, but this also prevents us from retiring older versions of packages such that we only have the most recent version. This matters if, for example, we wanted to migrate to a new repository location without backfilling every package ever.


imjasonh (approved):

xnox:
Is this kneejerk responses to:
- https://github.com/wolfi-dev/os/pull/31548
- https://github.com/wolfi-dev/os/pull/32245
- https://github.com/wolfi-dev/os/pull/32250

Which are all legit and now fixed?

xnox:
@jonjohnsonjr coached me on what this actually does and it is absolutely correct.

PR #1310, opened 2024-11-13
Enhance scriptability of advisory `create` and `update` commands

luhring:
## Note field

This adds a new `--note` flag that can be used to specify the `note` field of the new advisory event, for any event type that needs a note specified. This proposes that `--note` could eventually replace event-type-specific flags like `--fp-note` and `--tp-note`.

The following event types are now supported with `--note`: `pending-upstream-fix`, `analysis-not-planned`, `fix-not-planned`. For true and false positives, the type-specific flag takes priority over `--note`.

These type-specific flags have also been deprecated and will be removed in a future release.

## Disabling distro auto-detection

Previously, when distro auto-detection was disabled explicitly via the CLI flag, the `packageRepositoryURL` value wasn't checked ahead of time, so bogus requests were made for the APKINDEX and an opaque error was reported to the user.

Now, if distro auto-detection is disabled, all required values are checked individually, and any missing values (that won't be filled in because of said disabling) are reported to the user in an error.

Also, if auto-detection is not disabled, missing values are now filled in.

### DIFF
--- a/pkg/cli/advisory.go
+++ b/pkg/cli/advisory.go
@@ -94,7 +94,7 @@ func resolveTimestamp(ts string) (v2.Timestamp, error) {
 }
 
 type advisoryRequestParams struct {
-	packageName, vuln, eventType, truePositiveNote, falsePositiveNote, falsePositiveType, timestamp, fixedVersion string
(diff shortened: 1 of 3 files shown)


imjasonh (approved):

philroche:
Post merge +1. Thanks

PR #1350, opened 2024-12-03
fix(adv): validate that advisory IDs are CGA IDs

luhring:
Before, we were only validating that the ID was a valid _vulnerability_ ID, but not specifically a valid **_CGA_** ID.

#### Example

```console
$ w adv validate
❌ advisory data is not valid.

basic validation failure(s):
    grafana-10.3:
        advisories:
            CVE-2024-28180:
                "CVE-2024-28180" is not a valid CGA ID
```


cmdpdx (approved):

luhring:
Should've checked the other tests that were relying on the looser validation 🤦  I'll clean those up!

imjasonh (approved):

PR #1365, opened 2024-12-10
fix(osv): stop adding ID of record to "related" too

luhring:
The inclusion of the OSV record's ID as a related record was never necessary, and caused confusion downstream, such as on osv.dev.

### DIFF
--- a/pkg/advisory/osv.go
+++ b/pkg/advisory/osv.go
@@ -156,9 +156,9 @@ func BuildOSVDataset(ctx context.Context, opts OSVOptions) error {
 					continue
 				}
 
-				// Note: The OSV data should include our advisory ID itself among the listed
-				// related vulnerability IDs.
-				related := append([]string{adv.ID}, adv.Aliases...)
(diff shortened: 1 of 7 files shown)


luhring:
cc: @andrewpollock @oliverchang @cpanato

andrewpollock:
Will you retroactively correct existing records after this change?
(If yes, please ensure the `modified` field is updated also)

luhring:
> Will you retroactively correct existing records after this change? (If yes, please ensure the `modified` field is updated also)

Interesting, good callout. This is not accounted for in my change so far.

Today we take the latest "event" in our raw advisory data (e.g. [this timestamp](https://github.com/wolfi-dev/advisories/blob/5b9378754fa40538a5bc78bfde2bd8a836257369/bat.advisories.yaml#L23)) and use it as the modified time, which won't change by virtue of modifying how we produce the OSV data from the same (unmodified) raw advisory data. So we probably need some sort of "or X, whichever is later" logic. 🤔

PR #1372, opened 2024-12-13
Fail On Duplicate Config Names

paulgibert:
Fixes [#1599](https://github.com/chainguard-dev/internal-dev/issues/1599)

- Updated `wolfictl lint` to fail if a duplicate package config name is encountered.
- Added a unit test to `melange_test.go` for duplicate package config names

Steps to test:

1) See that `lint` succeeds on a valid repo:
```
git clone https://github.com/wolfi-dev/os.git
cd os
wolfictl lint -s error

# Should succeed
```

2) See that introducing duplicate configs results in an error:
```
# From inside os/

echo "package:\n  name: duplicate\n  version: 1.2.3" > duplicate1.yaml
cp duplicate1.yaml duplicate2.yaml
wolfictl lint -s error

# Example output: 2024/12/13 11:44:33 ERRO Package config names must be unique. Found duplicate 'duplicate'
```

The check happens in `pkg/melange` as this is where the duplicates were ignored as pointed out in [#1599](https://github.com/chainguard-dev/internal-dev/issues/1599). An alternative but more complex solution would be to add a new lint rule. I didn't go that route because it would require a notable ABI change.

#first-golang-pr (let's go)

@priyawadhwa


pkg/melange/testdata/duplicates_dir/foo2.yaml
```diff
@@ -0,0 +1,3 @@
+package:
+  name: foo
```
imjasonh, line 2: I think we might also want to fail if `.package.name` is not the same as the filename.

These are potentially different/separate issues, but it does I think violate some assumptions we've made, and is worth avoiding.

If we have that check, we also get uniqueness "for free" since there can't be two `foo.yaml`s

imjasonh (dismissed):

89luca89 (dismissed):

pkg/melange/testdata/duplicates_dir/foo2.yaml
paulgibert: If the .package.name field must match the filename, couldn't you make an argument that the field is redundant and should be removed?

pkg/melange/melange.go
```diff
@@ -143,8 +143,15 @@ func ReadAllPackagesFromRepo(ctx context.Context, dir string) (map[string]*Packa
...
 		}
-
-		p[packageConfig.Package.Name] = &Packages{
+		
+		name := packageConfig.Package.Name
+		
+		// check that the package config name is unique
+		_, exists := p[name]
+		if exists {
+			return p, fmt.Errorf("Package config names must be unique. Found duplicate '%s'", name)
```
priyawadhwa, line 152: It might be helpful to include the file names for the duplicate packages in the error message so it's easier to fix

paulgibert: Added file names to error message

pkg/melange/testdata/duplicates_dir/foo2.yaml
paulgibert: I added a check to error if the config name does not match the package name. This makes it impossible to test the package duplicate logic because, as you said, this gives us uniqueness for free, but I left the original check in just in case.

imjasonh: It probably is redundant, but I don't think we can easily remove it without causing lots and lots of downstream effects. I think it's fine to leave it, and just ensure it's the same as the filename, for now.

imjasonh (dismissed):

priyawadhwa (dismissed):

priyawadhwa (approved):

priyawadhwa:
Nice!

Issue #1220, opened 2024-10-02
wolfictl adv create truncates input line


wlynch:
**Description**

When creating an advisory with `wolfictl adv create`, if the `Note:` field exceeds the width of the terminal, the display truncates the message and you can no longer see what you are typing.

luhring:
This looks relevant 👀  https://github.com/charmbracelet/bubbles/issues/483

iamrajiv:
Hey @wlynch @luhring, I have fixed this issue and tested it locally by running wolfictl. Can you check my PR i.e. https://github.com/wolfi-dev/wolfictl/pull/1430?

PR #1430, opened 2025-02-03
fix(cli): Implement terminal wrapping for advisory note input

iamrajiv:
Fixes: #1220

Fixes an issue where the note field in `wolfictl adv create` would truncate text beyond terminal width, making long notes difficult to input and read. Implements automatic line wrapping that maintains visibility of the full text as users type.

### DIFF
--- a/pkg/cli/advisory.go
+++ b/pkg/cli/advisory.go
@@ -92,7 +92,7 @@ func resolveTimestamp(ts string) (v2.Timestamp, error) {
 	return v2.Timestamp(t), nil
 }
 
-// Add this function at the package level
+// getMultiLineInput is a helper function to get multi-line input from the user
 func getMultiLineInput(prompt string) (string, error) {
 	fmt.Print(prompt)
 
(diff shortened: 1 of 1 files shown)


iamrajiv:
Hi @Luhring, can you review this and run the CI?

rajiv-maersk:
@imjasonh

luhring (changes requested):
Hey, thanks for trying this out! And apologies for the delay in reviewing.

It looks like this approach is to prompt for the note upfront, before the existing prompting code gets invoked. It'd be great if we could incorporate the solution into the existing prompting flow, even if we swap out the component we're using today for the note field itself.

One reason this is needed in particular is that we don't always _want_ to prompt for the note — since not all advisory creations/updates end up needing a note field. We become aware of whether the note is needed throughout the existing prompt flow.

Check out the existing component models in `./pkg/cli/components/advisory`, and [this function](https://github.com/wolfi-dev/wolfictl/blob/26727365e56bdd4aa1e84d5e948c910597a0debc/pkg/cli/components/advisory/prompt/model.go#L282) in particular to understand how the note field relates to the rest of the UX.

PR #1436, opened 2025-02-08
feat(adv): `rebase` command and library functionality

luhring:
Adds a new command: `wolfictl adv rebase` — along with the accompanying library functionality.

This new command is useful for replaying the latest advisory event from one or more advisories in a "source" directory onto corresponding advisory data in a "destination" directory. `detection` events and `fixed` events are **not** copied. Newly added events use the current timestamp so they'll pass validation. Any net new advisories are given a newly generated CGA ID.

This is useful for when a package Melange file is copied from one repo to another, and the new instance of the package may have already accumulated some advisory data (e.g. `detection` events), but any conclusions from the original repo's advisory data should be applied to the new repo's advisory data as well.

### Syntax

```
wolfictl adv rebase <path-to-source-advisories-file.yaml> <path-to-destination-directory>
```

### Example

```
wolfictl adv rebase ./argo-cd-2.8.yaml ../enterprise-advisories
```

_**Note**: This PR also smooths out the `testerfs` implementation to make it easier to use filesystem test fixtures to observe whether code under test had the desired impact on the filesystem._


cpanato (approved):
lgtm

thanks

PR #1447, opened 2025-02-14
go-1.24: move to go 1.24

ajayk:


.github/workflows/golangci-lint.yaml
```diff
@@ -28,6 +28,6 @@ jobs:
           cache: false
 
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@051d91933864810ecd5e2ea2cfd98f6a5bca5347 # v6.3.2
+        uses: golangci/golangci-lint-action@0adbc47a5910e47adb692df88187ec8c73c76778 # v6.4.0
         with:
-          version: v1.63 # Please keep this in sync with the Makefile!
+          version: v1.64.5 # Please keep this in sync with the Makefile!
```
cpanato, line 33: just keep `1.64` with that it will pick all patchs releases and we dont need to update everytime

Makefile
```diff
@@ -75,7 +75,7 @@ golangci-lint:
 	set -e ;\
 
 	# Please keep the installed version in sync with .github/workflows/golangci-lint.yaml
-	GOBIN=$(GOLANGCI_LINT_DIR) go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.63.4 ;\
+	GOBIN=$(GOLANGCI_LINT_DIR) go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.5 ;\
```
cpanato, line 78: here is fine to use the full semver

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/wolfi-dev/wolfictl
 
-go 1.23.4
+go 1.24.0
```
cpanato, line 3: should we require for others consume this to be on go1.24?

.github/workflows/golangci-lint.yaml
luhring: One thing I like about specifying the full patch version here is that it ensures parity between what we run locally with `make lint` and what CI is running, so we don't get surprises in CI

go.mod
xnox: yes! also next syft release will bump us to 1.24 anyway.

xnox:
Resolved conflicts and addressed review comments in https://github.com/wolfi-dev/wolfictl/pull/1454

PR #1453, opened 2025-02-17
go.mod: set toolchain to go1.24.0 & update github actions

xnox:
Locally and in ci this will preffer, but not require, 1.24. As
currently we preffer to pick up 1.24 features (i.e. PQC TLS), but do
not explicitely require it.

Regularly bump the toolchain stanza to new upstream releases.

Whenever a toolchain feature becomes required do bump go stanza.

Confirmation that new workflow works correctly can be seen at:
- https://github.com/xnox/wolfictl/actions/runs/13380270071/job/37367480931

This is alternative to:
- https://github.com/wolfi-dev/wolfictl/pull/1447

Related:
- https://github.com/actions/setup-go/issues/457

### DIFF
--- a/.github/actions/action.yaml
+++ b/.github/actions/action.yaml
@@ -8,10 +8,12 @@ runs:
         repository: wolfi-dev/wolfictl
         path: wolfictl-setup-gha
 
+    # Until https://github.com/actions/setup-go/issues/457 is fixed
+    - id: gotoolchain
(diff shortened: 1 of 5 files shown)


xnox:
and this will all explode upon next syft upgrade that bumps to 1.24 see https://github.com/anchore/syft/pull/3660

PR #1474, opened 2025-03-03
Use Chainguard API to validate a generated CGA ID is unique

cmdpdx:
CGA IDs are intended to be unique per advisory. However, there have been a handful of advisories created with a preexisting CGA ID, usually when they are in different advisory repos. The advisory endpoints of the Chainguard API allow unauthenticated access to list advisory documents. We can use this to ensure that generated CGA IDs are in fact unique across all current advisories, regardless of originating repo.

### DIFF
--- a/pkg/advisory/cga_id.go
+++ b/pkg/advisory/cga_id.go
@@ -9,8 +9,8 @@ import (
 	advisoryapi "chainguard.dev/sdk/proto/platform/advisory/v1"
 )
 
-// ChainguardAPIURL is the public Chainguard API.
-const ChainguardAPIURL = "https://console-api.enforce.dev"
+// chainguardAPIURL is the public Chainguard API.
+const chainguardAPIURL = "https://console-api.enforce.dev"
 
 var DefaultIDGenerator IDGenerator = &RandomIDGenerator{}
 
@@ -67,7 +67,7 @@ func (s StaticIDGenerator) GenerateCGAID() (string, error) {
 	return s.ID, nil
 }
 
-type StaticListIDGenerator struct {
+type staticListIDGenerator struct {
 	// The set of IDs to return sequentially.
 	IDs []string
 
@@ -75,7 +75,7 @@ type StaticListIDGenerator struct {
 	idx int
 }
 
-func (s *StaticListIDGenerator) GenerateCGAID() (string, error) {
+func (s *staticListIDGenerator) GenerateCGAID() (string, error) {
 	if len(s.IDs) == 0 {
 		return "", fmt.Errorf("IDs list is empty")
 	}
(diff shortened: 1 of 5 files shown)


luhring:
I'll take a look at this tomorrow!

pkg/advisory/cga_id.go
```diff
@@ -60,3 +66,46 @@ type StaticIDGenerator struct {
 func (s StaticIDGenerator) GenerateCGAID() (string, error) {
 	return s.ID, nil
 }
+
+type StaticListIDGenerator struct {
```
luhring, line 70: Some of these new things are exported but not used in other packages in wolfictl — I'm guessing you need them for something downstream but just confirming!

pkg/advisory/create.go
```diff
@@ -88,9 +88,23 @@ func Create(ctx context.Context, req Request, opts CreateOptions) error {
...
+	// Continuously generate new CGA IDs until we get a unique one.
+	for {
+		id, err := GenerateCGAID()
+		if err != nil {
+			return fmt.Errorf("generating CGA ID: %w", err)
+		}
+
+		exists, err := CGAIDExists(ctx, id)
+		if err != nil {
+			return fmt.Errorf("checking existence of %s: %w", id, err)
```
luhring, line 102:
```suggestion
			return fmt.Errorf("checking existence of ID %q: %w", id, err)
```

(Just to be clear that we're checking on the ID, not the advisory named by that ID)

pkg/advisory/create_test.go
```diff
@@ -16,9 +18,30 @@ func TestCreate(t *testing.T) {
 	testTime := v2.Timestamp(time.Date(2022, 9, 26, 0, 0, 0, 0, time.UTC))
 	brotliExistingEventTime := v2.Timestamp(time.Date(2022, 9, 15, 2, 40, 18, 0, time.UTC))
 
+	goodCGAID := "CGA-xoxo-xoxo-xoxo"
+	usedCGAID := "CGA-abcd-abcd-abcd"
```
luhring, line 22: nit: It might help us with sanity if this used ID was invalid for _only_ the reason that it's used — as opposed to also because it includes invalid characters:

```go
allowedChars := "23456789cfghjmpqrvwx"
```

But curious for your thoughts

luhring (commented):
Nice — I think the quality of the code you've added is great.

I just have to double check the impetus — are we sure we saw duplicate IDs being generated by this same function? I know we've had issues with CGA IDs being reused, because of bad logic or human error where we're copying advisories wholesale, say across our various repos. But assuming we're handling randomness correctly, it seems virtually impossible that we'd actually generate the same ID twice. I just checked with my friend ChatGPT who said:

>In other words, there’s roughly a 1 in 117 billion chance of a collision on any given new ID generation.

So if it's possible to skip the network call for an otherwise very fast operation, that'd be ideal. But if you've seen evidence the function is misbehaving, I agree we need to address that!

cmdpdx:
@luhring you caught this before I was able to post about it! 😄 

> are we sure we saw duplicate IDs being generated by this same function?

I have three crons running hourly ingesting the current advisory repos into datastore. The CGA ID is used as the primary key, and I've received a handful of collision errors. That was the motivation to begin this change.

In parallel, I was also investigating the collisions to confirm that's what I was seeing. It turned out a few of them were due to package names on an advisory being updated, rather than a duplicate ID generated. Those will need to be fixed on the ingestion end in a different way.

However, I was able to find two cases of duplicate IDs used across different advisory repos for distinct packages. I will post those details in the sustaining channel for follow up. 

The network call is fairly quick, but if it becomes an issue I think it would make sense to add an endpoint to handle querying single advisories by CGA ID, rather than leveraging the existing `ListDocuments` which uses the CGA ID in the list of CVE IDs to filter by.

pkg/advisory/cga_id.go
cmdpdx: Nope good call out, I was mostly following some of the given examples but currently nothing outside of `advisory` needs anything I've added. I unexported it all unless/until we determine they can be used elsewhere.

pkg/advisory/create_test.go
cmdpdx: I'm all in favor of readability and doing our future selves favors. Updated to only use valid chars.

luhring:
Just replied in the Slack thread! **tl;dr:** I'm not sure this PR is needed, but we can keep it open for now while we discuss!

cmdpdx:
Closing in favor of doing this check in CI

PR #1475, opened 2025-03-04
feat(sbom,scan): leverage CPE data for Go module matching

luhring:
With a goal of finding more true positives for the Go ecosystem, this PR turns on CPE matching for Go modules.

With this, we are also investing in better CPE assignments during SBOM generation, particularly for the `golang.org/x/...` group of repositories.

Also, since we're still using some Syft-generated CPEs, we've added a step at the end of matching to filter out CPE-based matches that don't meet our standards.

⚠️ Like all changes of this nature, it's possible this will produce unexpected false positives/negatives in our APK scans. However, since we're in control of everything we've just added, we should be able to quickly respond to any behavioral changes we don't like.

### DIFF
--- a/pkg/sbom/testdata/goldenfiles/aarch64/terraform-1.5.7-r12.apk.syft.json
+++ b/pkg/sbom/testdata/goldenfiles/aarch64/terraform-1.5.7-r12.apk.syft.json
@@ -4904,8 +4904,24 @@
       "language": "go",
       "cpes": [
         {
-          "cpe": "cpe:2.3:a:golang:x\\/crypto:v0.21.0:*:*:*:*:*:*:*",
-          "source": "syft-generated"
+          "cpe": "cpe:2.3:a:golang:crypto:v0.21.0:*:*:*:*:go:*:*",
+          "source": "wolfictl"
+        },
+        {
+          "cpe": "cpe:2.3:a:golang:bcrypt:v0.21.0:*:*:*:*:go:*:*",
+          "source": "wolfictl"
+        },
+        {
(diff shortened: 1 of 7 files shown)


pkg/sbom/testdata/goldenfiles/aarch64/terraform-1.5.7-r12.apk.syft.json
```diff
@@ -4904,8 +4904,24 @@
       "language": "go",
       "cpes": [
         {
-          "cpe": "cpe:2.3:a:golang:x\\/crypto:v0.21.0:*:*:*:*:*:*:*",
-          "source": "syft-generated"
```
luhring, line 4908: This is a neat example of how we'll be cleaning up bad CPEs from our SBOMs 🤓

imjasonh (approved):

pkg/scan/apk.go
```diff
@@ -250,6 +273,64 @@ func (s *Scanner) APKSBOM(ctx context.Context, ssbom *sbomSyft.SBOM) (*Result, e
 	return result, nil
 }
 
+// shouldAllowMatch is a point where we can optionally filter out matches from
+// the scan based on criteria we define. This function will return true unless
+// it determines that the match should be dropped from the final result set. In
+// this latter case, it also returns a string explanation of the reason for
+// dropping the match.
+func shouldAllowMatch(m match.Match) (allow bool, reason string) {
```
vaikas, line 281: Could you add a unit test for this?

luhring: Yes — coming right up

luhring: https://github.com/wolfi-dev/wolfictl/pull/1476

PR #1495, opened 2025-03-14
feat(scan/apk): Respect GRYPE_DB_MAX_ALLOWED_BUILT_AGE

EyeCantCU:
There was about a one hour window of time where I couldn't scan Hadoop or gRPC while working through vulnerabilities without possibly building the database locally (because it hadn't been updated yet) so I took a quick look at how we handle this in wolfictl and found it deviates from the behavior in grype

Similar to grype, respect GRYPE_DB_MAX_ALLOWED_BUILT_AGE so that users can set this to whatever their preferred duration is. Default to 24 hours


pkg/scan/apk.go
```diff
@@ -139,11 +139,16 @@ func NewScanner(opts Options) (*Scanner, error) {
...
+	if err != nil {
+		grypeMaxAllowedBuiltAge = time.Duration(24)
+	}
+
 	installCfg := installation.Config{
 		DBRootDir:               dbDestDir,
 		ValidateChecksum:        true,
 		ValidateAge:             !opts.DisableDatabaseAgeValidation,
-		MaxAllowedBuiltAge:      24 * time.Hour,
+		MaxAllowedBuiltAge:      grypeMaxAllowedBuiltAge * time.Hour,
```
developer-guy, line 151: to fix `Multiplication of durations:` here, it think we should do:

```go
MaxAllowedBuiltAge:  time.Duration(grypeMaxAllowedBuiltAge) * time.Hour,
```

EyeCantCU: `grypeMaxAllowedBuiltAge` already shared the type. I was multiplying two durations which is wrong. Should be sorted now. Thank you for taking a look!

hectorj2f (changes requested):
Could we add this as a flag? So this doesn't get hidden into the code. That way we could add some documentation about it.

PR #1511, opened 2025-03-19
bump grype to v0.89.1 to fix some issues

hectorj2f:
We are facing this issue in our builds https://github.com/anchore/grype/pull/2518 because we are using an old version of Grype.

related to this: https://github.com/chainguard-dev/internal-dev/issues/10619


hectorj2f:
To upgrade to latest version, we might need to make some additional code changes. Those can be done in a different PR.

luhring (changes requested):
Looking into it — let's hold off for just a short bit for now — I'll explain I promise!

PR #1512, opened 2025-03-19
config(scan/apk): disable maven searches until we resolve the 403 rate limit issues with the newer release of Grype

tdunlap607:
This PR temporarily disables Maven searches for java-archive packages. The reason for this is because in [v0.89.1 of Grype](https://github.com/anchore/grype/releases/tag/v0.89.1) these 403s rate limits are now surfaced as fatal errors, which would cause our scanning to fail as well.


philroche (approved):
Based on our goal for consistency while we work to re-introduce maven central lookups in a deterministic consistent way +1 on this proposal.

luhring (approved):

PR #1518, opened 2025-03-28
feat(adv): bulk create and update of adv data

luhring:
This updates the `wolfictl adv create` and `... update` commands to be able to handle updates of multiple packages and/or vulnerabilities in one fell swoop — both via the CLI flags and the interactive prompts.

You can specify more than one item for both packages and vulnerabilities using a comma-separated list, e.g.:

```console
wolfictl adv create -p apko,age -V CVE-2025-1111,CVE-2025-1112 -t fix-not-planned --note 'insert reason'
```

`wolfictl` handles this by applying the given advisory event parameters to the cross product of any listed packages and vulnerabilities. So in the above example, we'll create a new advisory with the `fix-not-planned` event, for two packages, and two vulnerabilities for each of those two packages — resulting in the creation of **4 new advisories total**.

Less interesting things this PR also does:

- Removes `detection` and `analysis-not-planned` from the list of selectable event types from the interactive prompt (they shouldn't be needed)
- Uses the new readonly `Getter` abstraction in places where the `create` and `update` command logic have use for readonly advisory access
- Allows interactive prompt models to return an error to the user
- Adds a new handy `CompleteAliasSet` function for finding all aliases without using the heavier `DiscoverAliases` logic which depends on a `configs.Index[v2.Document]` instance
- Promotes the "Advisory Request Params" concept from a private type in the `cli` package to a public type in the `advisory` package, to allow for the separation of data gathering (including via e.g. interactive prompts) and data access (i.e. the `Request` type passed to `advisory.Create` and `advisory.Update`).

### DIFF
--- a/pkg/advisory/store.go
+++ b/pkg/advisory/store.go
@@ -24,3 +24,7 @@ type Getter interface {
 	// valid, non-empty data.
 	Advisories(ctx context.Context, packageName string) ([]v2.PackageAdvisory, error)
 }
+
+type Putter interface {
+	Upsert(ctx context.Context, request Request) (string, error)
+}
--- a/docs/cmd/wolfictl_advisory_create.md
+++ b/docs/cmd/wolfictl_advisory_create.md
@@ -28,6 +28,10 @@ command line, the command will prompt for the missing values.
 If the --no-prompt flag is specified, then the command will fail if any
 required fields are missing.
 
+It's possible to create advisories for multiple packages and/or vulnerabilities 
+at once by using a comma-separated list of package names and vulnerabilities. 
+This is available for both the CLI flags and the interactive prompt fields.
+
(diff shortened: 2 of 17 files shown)


pkg/advisory/store.go
```diff
@@ -24,3 +24,7 @@ type Getter interface {
 	// valid, non-empty data.
 	Advisories(ctx context.Context, packageName string) ([]v2.PackageAdvisory, error)
 }
+
+type Putter interface {
```
vaikas, line 28: Are we also going to have a PitchingWedge and Driver at some point?

vaikas (approved):

pkg/advisory/store.go
luhring: Yes!

PR #1525, opened 2025-04-02
feat(adv): on bulk operations for `create` and `update`, skip event additions that would be redundant by event type

luhring:
This was requested by the Sustaining team, to help ensure that if they're operating en masse on a set of advisories, bulk operations are resilient to the scenario where one or two of the advisories in the target set have already received their updates via some other mechanism. The command logs a warning when this happens, so the user knows that part of the requested update was skipped, in case they want to add an event to the skipped advisory after all.

This PR also catches up on the auto-gen'ed documentation.

### DIFF
--- a/pkg/advisory/request.go
+++ b/pkg/advisory/request.go
@@ -320,15 +320,15 @@ func resolveTimestamp(ts string) (v2.Timestamp, error) {
 	return v2.Timestamp(t), nil
 }
 
-// MatchPackageAdvisoryToRequest takes an input slice of PackageAdvisory and a
-// Request and returns the first PackageAdvisory that matches the Request. If no
-// match is found, it returns nil.
+// MatchToRequest takes an input slice of PackageAdvisory and a Request and
+// returns the first PackageAdvisory that matches the Request. If no match is
+// found, it returns nil.
 //
 // A "match" is defined as meeting all the following criteria: having the same
 // Package name; if the Request's AdvisoryID is set, it must match the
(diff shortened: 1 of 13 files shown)


pkg/advisory/request.go
```diff
@@ -319,3 +319,37 @@ func resolveTimestamp(ts string) (v2.Timestamp, error) {
...
+
+// MatchPackageAdvisoryToRequest takes an input slice of PackageAdvisory and a
+// Request and returns the first PackageAdvisory that matches the Request. If no
+// match is found, it returns nil.
+//
+// A "match" is defined as meeting all the following criteria: having the same
+// Package name; if the Request's AdvisoryID is set, it must match the
+// PackageAdvisory's ID; and if the Request has Aliases and no AdvisoryID, at
+// least one of the Aliases must match the PackageAdvisory's Aliases.
+func MatchPackageAdvisoryToRequest(advs []v2.PackageAdvisory, req Request) *v2.PackageAdvisory {
```
shimmerjs, line 331: Food for thought: since the package is already named `advisory`, it implies part of the name, eg `advisory.MatchToRequest(...)`

shimmerjs (dismissed):
sleigh, queen

pkg/advisory/request.go
luhring: On it

luhring: Done

shimmerjs (approved):

PR #1539, opened 2025-04-11
fix(scan): increase allowed age for Grype DB

luhring:
When Anchore fails to build a DB on a given day, the 24 hour max age becomes too small for our ability to scan successfully. Increase to 48 hours to allow for more resilience without risking silently scanning with overly stale data.


cpanato (approved):

cpanato:
can we make that configurable?

luhring:
> can we make that configurable?

Yeah I'd be fine with that

PR #1540, opened 2025-04-11
refactor(scan): update scan filtering to use new `advisory.Getter` abstraction

luhring:
This also introduces a nifty mechanism for taking an existing, old-style `*configs.Index[v2.Document]` and adapting it to the new `advisory.Getter` interface so that consuming code can move ahead to the abstraction-based model when possible.

### DIFF
--- a/pkg/advisory/index_adapter.go
+++ b/pkg/advisory/index_adapter.go
@@ -1,0 +1,58 @@
+package advisory
+
+import (
+	"context"
+	"fmt"
+	"sort"
+
+	"github.com/wolfi-dev/wolfictl/pkg/configs"
+	v2 "github.com/wolfi-dev/wolfictl/pkg/configs/advisory/v2"
+)
+
+type indexAdapter struct {
+	index *configs.Index[v2.Document]
+}
+
+// AdaptIndex creates an implementation of advisory.Getter using an existing
+// instance of `*configs.Index[v2.Document]`.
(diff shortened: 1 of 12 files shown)


pkg/advisory/index_adapter.go
```diff
@@ -0,0 +1,58 @@
+package advisory
```
vaikas, line 1: Looks like we don't use boilerplate in these repo, and when we do it's interesting. No action necessary, just an observation.

pkg/advisory/store.go
```diff
@@ -54,3 +54,20 @@ type Putter interface {
 	// returned.
 	Upsert(ctx context.Context, request Request) (string, error)
 }
+
+// MapByVulnID maps the given advisories by their vulnerability ID, creating a
```
vaikas, line 58: This seems like an implementation detail that does not belong in store.go (which I think should be used for the interface definition).

luhring: That sounds good, I can move it somewhere else

pkg/advisory/index_adapter.go
luhring: 👍

pkg/advisory/store.go
luhring: Moved to a new file

vaikas (approved):

PR #1548, opened 2025-04-16
adv: introduce the new advisory-schema module

mgreau:
This commit introduces the new [`advisory-schema` module ](https://github.com/chainguard-dev/advisory-schema)which contains the advisory schema and validation logic extracted from wolfictl.

It ensures a clear separation between the codebase that defines the schema and the codebases that consumes it, such as this `wolfictl` repository and others.

Note - the initial release version of `advisory-schema` (v0.37.x) is aligned with the upcoming `wolfictl` release that first depends on this module, making the refactoring history easier to trace. Each project will  follow its own release schedule moving forward.


cmdpdx (approved):
LGTM, though if possible I'd like to have @luhring puts eyes on it just in case I missed something nuanced.

Also, this should wait until the datastore changes land, since otherwise we still depend on the removed code here.

pkg/advisory/v2/index.go
luhring, line 1: Did this stuff have to move? I was thinking we'd keep this in `pkg/configs/v2`. So that would be our YAML-based access code, it just wouldn't have any schema definition any more.

luhring (approved):

PR #1569, opened 2025-04-29
bump: Don't fail if there are comments

jonjohnsonjr:
See https://github.com/wolfi-dev/os/pull/52024


dannf (approved):
nice.

kinda seems odd that we continue pattern matching even once we've bumped the epoch. but that's a pre-existing thing.

PR #1575, opened 2025-05-05
bug(scan/filter): filtering considered all advisories and not just the target package

tdunlap607:
We noticed that filtering was accounting for all advisories, which unintentionally reduced the original result scan set.

For example, simply adding a new advisory file (e.g., under a new package name like `foo.advisories.yaml`) with matching vulnerability aliases would cause existing tests to fail. This happened because `advGetter.PackageNames` was collecting all advisories (e.g., from wolfi-dev/advisories) and applying them during filtering.

However, the correct behavior appears to be filter only using advisories that match the incoming `result.TargetAPK.Name` OR `result.TargetAPK.OriginPackageName` (to account for subpackages).

@luhring thoughts?


luhring (approved):
Magnificent!

PR #1577, opened 2025-05-06
fix: return ErrNoEntries and capture it

hectorj2f:
This is related to one issue with the scanning of apks when it should have warned and continued. The IndexAdapter was returning a custom error when no advisory document were found for a package. However other advisory getters such as FSGetter was propagating the **ErrNoEntries** which was then captured in the calling functions. 
We must capture that error and return the same findings given in the function call when detected an **ErrNoEntries**.

### DIFF
--- a/pkg/advisory/index_adapter.go
+++ b/pkg/advisory/index_adapter.go
@@ -2,10 +2,11 @@ package advisory
 
 import (
 	"context"
-	"fmt"
+	"errors"
 	"sort"
 
 	v2 "github.com/chainguard-dev/advisory-schema/pkg/advisory/v2"
+	"github.com/chainguard-dev/clog"
 	"github.com/wolfi-dev/wolfictl/pkg/configs"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/scan/filter.go
```diff
@@ -38,6 +40,10 @@ func FilterWithAdvisories(ctx context.Context, result Result, advGetter advisory
 	// Check for advisories in the result target's origin
 	packageAdvisories, err := advGetter.Advisories(ctx, result.TargetAPK.Origin())
 	if err != nil {
+		if errors.Is(err, configs.ErrNoEntries) {
+			log.Warn("no source advisories found for package, skipping")
+			return filteredFindings, nil
+		}
```
luhring, line 46: (This becomes unnecessary per my other comment re: `ErrNoEntries` — but we could instead check for `len(packageAdvisories) == 0`)

pkg/advisory/index_adapter.go
```diff
@@ -38,7 +39,11 @@ func (c indexAdapter) PackageNames(_ context.Context) ([]string, error) {
 func (c indexAdapter) Advisories(_ context.Context, packageName string) ([]v2.PackageAdvisory, error) {
 	entry, err := c.index.Select().WhereName(packageName).First()
 	if err != nil {
-		return nil, fmt.Errorf("getting advisory document for %q: %w", packageName, err)
+		if errors.Is(err, configs.ErrNoEntries) {
+			log.Warnf("no source advisories found for package %q, skipping", packageName)
```
luhring, line 43: Sorry for the nit — can we use clog here? Can use the `context.Context` passed into this `Advisories(...)` func.

Just so this fits well in our internal logging setup

luhring (approved):

PR #1601, opened 2025-05-16
feat(adv): new FSPutter constructor for automatic encode config

luhring:
Adds a new constructor, `advisory.NewFSPutterWithAutomaticEncoder(fsys rwfs.FS)`, as an easier way to create a fully-functional `FSPutter`. This avoids having to set up the yam encoder yourself, because it's done automatically for you on a best-effort basis, reading the `.yam.yaml` file from the FS you provide.


vaikas (approved):
Sweet!

PR #1606, opened 2025-05-19
Improve release workflow

cpanato:
Improve release workflow and also add the wolfictl binaries to the github release


luhring:
nit: if you could describe (in the PR description) just a little bit about what the new way to interact with this release process is, that'd be super helpful!

PR #1649, opened 2025-06-17
feat: Add linter to check for potentially hanging tests

antitree:
During the QEMU migration, we noticed that packages that have a pattern like so:

`        hostpathplugin --v=5 --endpoint="unix:///csi/csi.sock" --nodeid="node-000000" & sleep 5`

Create a potential for an infinite hang during tests. This linter provides a WARNING if that is detected to give users a hint in the case this happens to fail a build.


pkg/lint/rules.go
```diff
@@ -43,6 +43,9 @@ var (
 	hostEditDistanceExceptions = map[string]string{
 		"www.libssh.org": "www.libssh2.org",
 	}
+
+	// Detect background processes (commands ending with '&' or '& sleep ...') that do not redirect output
+	reBackgroundProcess = regexp.MustCompile(`&(?:\s*$|\s+sleep\b)`) // matches 'cmd &' or 'cmd & sleep'
```
egibs, line 48: I was working on a PR in Melange with a few additional patterns:
```go
daemonFlags := []string{
	`-d\b`,
	`--daemon\b`,
	`--daemonize\b`,
	`--detach\b`,
	`-daemon\b`,
}
daemonPattern := `.*(?:` + strings.Join(daemonFlags, "|") + `).*`

redirPatterns := []string{
	`>\s*\S+`,
	`>>\s*\S+`,
	`2>\s*\S+`,
	`2>>\s*\S+`,
	`&>\s*\S+`,
	`&>>\s*\S+`,
	`>\s*\S+.*2>&1`,
	`2>&1.*>\s*\S+`,
	`>\s*/dev/null`,
	`2>\s*/dev/null`,
	`&>\s*/dev/null`,
	`\d+>&\d+`,
}
```

There are also examples like:
```sh
/entrypoint foo:pass:::upload & wait-for-it localhost:22
```
and
```
sudo -E -u user /usr/share/foo/bin/bar -d \
  ... \
  ... \
  -p /tmp/foo.pid
```


The latter runs in the background and is not obvious. The former is difficult since it's nestled in the middle of the line.

antitree:
Factored in your changes @egibs and did some reviews for false positives. Based on the wolfi-dev/os from 30 days ago, it catches many of the issues.

stevebeattie:
This is good stuff, but I'm not sure how much the wolfictl linter is actively developed and is preferred; it seems more work on package source linting is done in melange's pkg/linter/ itself.

justinvreeland:
> package source linting is done in melange's pkg/linter/ itself.

I don't think melange linter does any package source (yaml file) linting (though it probably should). At least looking a the melange I have installed now it only lists things that would impact built packages.

PR #1652, opened 2025-06-18
chore(scan): add only primary evidence locations to findings

luhring:
We've decided to discard other locations from the list, in part to keep the list of locations per finding minimal and thus easier to reason about as we're examining vulnerabilities for a package over time.

### DIFF
--- a/pkg/scan/finding.go
+++ b/pkg/scan/finding.go
@@ -6,9 +6,7 @@ import (
 
 	"github.com/anchore/grype/grype/match"
 	"github.com/anchore/grype/grype/vulnerability"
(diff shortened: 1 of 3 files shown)


vaikas:
> Should we warn when skipping supporting locations? Also I thought Ville mentioned to keep these supporting locations too.

I think this is a good improvement, and we can add a "supporting locations" as an additional array, or something as a follow up to this, but I think this is an improvement in itself and lumping them all together is not really useful.

vaikas (approved):

hectorj2f (approved):

PR #1686, opened 2025-07-21
test(linter): reduce error messages ambiguity

konradzapalowicz:
The assert.Equal() expects the parameters to follow the "expected, actual" order. However, the test assertions for linter rules were written in the reversed order. While doing so has no functional consequences the output was confusing.

This commit fixes that by introducing the proper order of parameters.

### DIFF
--- a/pkg/lint/rules_test.go
+++ b/pkg/lint/rules_test.go
@@ -582,9 +582,9 @@ func TestLinter_Rules(t *testing.T) {
 			require.Len(t, g.Errors, len(tt.want.Errors))
 
 			for i, e := range g.Errors {
-				assert.Equal(t, e.Error, tt.want.Errors[i].Error, "Lint(): Error: got = %v, want %v", e.Error, tt.want.Errors[i].Error)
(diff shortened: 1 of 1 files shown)


konradzapalowicz:
For the record, this is how this has manifested itself:
```                
Error Trace:    /Users/konrad.zapalowicz/work/wolfi-dev/wolfictl/pkg/lint/rules_test.go:613
Error:          Not equal:
                   expected: &errors.errorString{s:"[update-identifier-must-match-git-repository]: update identifier does not match the repository URI (ERROR)"}
                   actual  : &errors.errorString{s:"[auto-update-identifier-matches-repo]: update identifier does not match the repository URI (ERROR)"}

Test:           TestLinter_Rules/update-identifier-not-matching-git-checkout-repository.yaml
Messages:       Lint(): Error: got = [update-identifier-must-match-git-repository]: update identifier does not match the repository URI (ERROR), want [auto-update-identifier-matches-repo]: update identifier does not match the repository URI (ERROR)
```

PR #1702, opened 2025-07-30
Revert yam

dannf:


kranurag7 (changes requested):
holding until we've more confirmation on yam issue.

PR #1716, opened 2025-08-11
withdraw: add option to read from file

javacruft:
As part of the work to garbage collect older apk files we need to be
able to withdraw more packages that can be supported as command line
arguments from APK indexes; add option to pass the list of packages to
withdraw as a file to support larger withdrawal options.

### DIFF
--- a/docs/cmd/wolfictl_withdraw.md
+++ b/docs/cmd/wolfictl_withdraw.md
@@ -19,9 +19,9 @@ withdraw --signing-key ./foo.rsa example-pkg-1.2.3-r4 also-bad-2.3.4-r1 <old/APK
 ### Options
 
 ```
-  -h, --help                        help for withdraw
-      --signing-key string          the signing key to use (default "melange.rsa")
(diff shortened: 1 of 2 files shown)


sil2100 (dismissed):
This looks good. I personally consider the `--withdrawn-packages` argument a bit unambiguous as I don't know from first glance if this means giving a filename or just a list, but I guess this is more what other tools do as well.

javacruft:
> This looks good. I personally consider the `--withdrawn-packages` argument a bit unambiguous as I don't know from first glance if this means giving a filename or just a list, but I guess this is more what other tools do as well.

I agreed - I went with --withdrawn-packages-file first and then dropped the suffix - do you think that makes it more obvious?

sil2100:
My only problem with `--withdrawn-packages-file` is the length. However...! How about we do `--packages-file` instead? The withdraw part is already in the command, so I think saying 'here is the packages file' doesn't require additional context? WDYT?

PR #1935, opened 2026-04-09
wolfictl bump - drop all epoch-line comments by default.

smoser:
The vast majority of comments alongside an epoch bump are redundant with 'git log --oneline' or at very least are not relevant after the next epoch bump.


sergiodj (approved):
LGTM, thanks.

zmarano (approved):

PR #1988, opened 2026-06-02
fix(dag): preserve provider-priority for local virtual providers

iainlane:
## Problem

Package builders order package builds using the dependency graph produced by `pkg/dag`. That graph combines a synthetic local APK repository (the unbuilt configs) with the remote APK indexes named by the melange configs. For the ordering to be correct, apk's resolver has to choose the same provider from the synthetic repo that it would choose from a real `APKINDEX`. If it chooses a remote package instead, the edge points at an external package and is dropped when the graph is filtered to local packages, leaving the two packages unordered and free to build in parallel.

This went wrong for virtual dependencies that use `provider-priority`. The synthetic local packages did not copy `provider-priority` from the melange configuration, so every local provider defaulted to priority `0`. When an already-published remote package provided the same virtual dependency at a higher priority, the resolver preferred the remote package. Its edge was then filtered out, and the dependent package built against the published artifact rather than the local rebuild.

Here is a case that triggers it. A `*-dev` subpackage provides a virtual `<foo>-dev` with a non-zero `provider-priority`, and another package build-depends on `<foo>-dev`:

```yaml
# foo.yaml
package:
  name: foo
  version: "2.0.0"
  epoch: 0
# ...
subpackages:
  - name: foo-py3.13-dev
    dependencies:
      provides:
        - foo-dev
      provider-priority: 313
```

```yaml
# bar.yaml (reverse dependency)
package:
  name: bar
  version: "1.0.0"
  epoch: 0
environment:
  contents:
    packages:
      - foo-dev   # build against the local rebuild of foo
# ...
```

When `foo` is bumped so that it ships `so:libfoo.so.2` in place of `so:libfoo.so.1`, `bar` should be ordered after the local `foo` rebuild and link against the new library. Before this change the local `foo-py3.13-dev` advertised `foo-dev` at priority `0`, while the previously published `foo` in the index advertised it at `313`. The resolver picked the remote provider, the edge was filtered out, and `bar` built in parallel against `so:libfoo.so.1`.

## Fix

Copy `provider-priority` onto the synthetic APK packages for both origin packages and subpackages, so a local provider ties its remote counterpart on priority and then wins on version. `provider-priority` is parsed the way melange permits it, as a signed integer: a negative value resolves to the default priority of `0` in the built index (`unbound-mailcow-compat` and `debezium-3.5` use `-1`), so it is treated as `0` here rather than aborting graph construction, while a non-numeric value is still reported as an error.

The first commit removes an unused per-origin resolver. It was stored in the resolver map under a `name-version` key but never looked up, because reads only use the comma-joined repository-source key. Subpackage-to-origin ordering is added directly with `AddEdge`. Removing it first means the fix does not have to return an error from code that has no effect.

## Tests

Adds regression tests for provider-priority preservation on a subpackage provider and for the parsing helper.

### DIFF
--- a/pkg/dag/graph.go
+++ b/pkg/dag/graph.go
@@ -13,7 +13,6 @@ import (
 	"path"
 	"reflect"
 	"sort"
-	"strconv"
 	"strings"
 
 	"github.com/dominikbraun/graph"
@@ -104,8 +103,6 @@ func NewGraph(ctx context.Context, pkgs *Packages, options ...GraphOptions) (*Gr
 			errs = append(errs, err)
 			continue
 		}
-		// add the origin package as its own resolver, so that the subpackage can resolve to it
-		g.resolvers[c.String()] = singlePackageResolver(ctx, c, opts.arch)
 		for i := range c.Subpackages {
 			subpkg := pkgs.Config(c.Subpackages[i].Name, false)
 			for _, subpkgVersion := range subpkg {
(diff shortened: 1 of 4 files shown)


xnox:
> ci / wolfictl text on wolfi-dev/os (pull_request) Failing after 2m

I think this check was dropped in production; and shouldn't be used to gate things anymore.

It also likely means we have build cycles too now; not sure if we care.

iainlane:
> > ci / wolfictl text on wolfi-dev/os (pull_request) Failing after 2m
> 
> I think this check was dropped in production; and shouldn't be used to gate things anymore.
> 
> It also likely means we have build cycles too now; not sure if we care.

Hmm, really? Well, it was a useful failure to me. It showed me that we'd start refusing to parse packages we actually have. There are two with `-1` there. So I did make the check a bit more permissive...

xnox:
> > > ci / wolfictl text on wolfi-dev/os (pull_request) Failing after 2m
> > 
> > 
> > I think this check was dropped in production; and shouldn't be used to gate things anymore.
> > It also likely means we have build cycles too now; not sure if we care.
> 
> Hmm, really? Well, it was a useful failure to me. It showed me that we'd start refusing to parse packages we actually have. There are two with `-1` there. So I did make the check a bit more permissive...

ouch, i thought -1 is invalid and not allowed.

justinvreeland:
> ouch, i thought -1 is invalid and not allowed.
it shouldn't be but the melange schema for yaml files is not as good as it should be

justinvreeland (approved):

iainlane:
@justinvreeland I don't seem to have access I'm afraid. Any chance you can take care of the merge please? :pray:

